PR #3, opened 2025-01-21
refactor: integrate with agglayer

Freyskeyd:
# Description

This PR is moving some deps around and refactor the agglayer-prover-types to not depends on `agglayer-types`.

## PR Checklist:


crates/agglayer-prover-types/src/generated/agglayer.prover.v1.rs
```diff
@@ -320,8 +376,8 @@ pub mod proof_generation_service_server {
         }
     }
     /// Generated gRPC service name
-    pub const SERVICE_NAME: &str = "agglayer.prover.v1.ProofGenerationService";
-    impl<T> tonic::server::NamedService for ProofGenerationServiceServer<T> {
+    pub const SERVICE_NAME: &str = "agglayer.prover.v1.PessimisticProofService";
```
atanmarko, line 379: Nit - should this be `kebab-case`?

Freyskeyd: This is generated code from `proto` files, and the [guideline](https://protobuf.dev/programming-guides/style/#services) defines that service names should be PascalCase :)

atanmarko (approved):

Ekleog-Polygon:
My review of this PR is currently blocked on a quick 1:1 to understand what exactly the goal of splitting the provers repo is and what the semantics are. Hopefully I’ll be able to review by EOW :)

crates/agglayer-prover-types/build.rs
```diff
@@ -1,11 +1,22 @@
 use std::path::PathBuf;
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let descriptor_path = PathBuf::from("src/generated").join("agglayer.prover.bin");
+    let current_package = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    let root = current_package.join("../../").canonicalize()?;
+    let proto_path = root.join("proto/agglayer/prover");
+    let descriptor_path = current_package
+        .join("src/generated")
+        .join("agglayer.prover.bin");
```
Ekleog-Polygon, line 9: We should move this to `OUT_DIR`, so it lives out of source code, gets rebuilt on each `cargo build`, and doesn’t need to be committed / to merge conflict all the time — just like [tonic does by default](https://docs.rs/tonic-build/0.12.3/tonic_build/struct.Config.html#method.out_dir)

I’ll probably do that as a follow-up PR if noone else takes it up before, considering this is not a regression.

crates/agglayer-prover-types/build.rs
```diff
@@ -1,11 +1,22 @@
...
+
+    let proof_generation_proto = proto_path.join("v1/proof_generation.proto");
 
     tonic_build::configure()
+        .type_attribute(
+            "GenerateProofError",
+            "#[derive(serde::Serialize, serde::Deserialize)]",
+        )
         .file_descriptor_set_path(descriptor_path)
         .out_dir("src/generated")
```
Ekleog-Polygon, line 19: Just like here, we could just use the default `OUT_DIR` to not have to always commit the files and merge conflict each time we change one of the source files of these… no?

Ekleog-Polygon (commented):
LGTM! Not approving yet due to the comments @hadjiszs already wrote :)

crates/agglayer-prover-types/build.rs
Freyskeyd: This is intentional that we're committing the generated file, the goal is:

- To be able to see generated code structs directly in source control, it is easier to debug and ease the integration with every tool.
- Avoid unwanted changes that can be missed if we do not have a diff on those file
- Make `protoc` not needed for some cases (mainly for CI, this is still in todo, I didn't have time to do it)

A better evolution would be to completely remove `tonic-build` and rely on `buf` gen, we'll see that later when we tackle some refactor.

Ekleog-Polygon (approved):

hadjiszs (approved):

atanmarko (dismissed):

crates/agglayer-prover-types/src/lib.rs
```diff
@@ -34,3 +36,86 @@ pub enum Error {
...
+                (tonic::Code::InvalidArgument, value.to_string(), details)
+            }
+        };
+
+        Ok(Status::with_details(code, message, details.into()))
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, thiserror::Error, PartialEq, Eq)]
+pub enum ProofVerificationError {
```
iljakuklic, line 87: Nit: would be nice to put this (and related code) into an `error` submodule

iljakuklic (approved):

atanmarko (approved):

PR #14, opened 2025-01-23
feat: adding aggkit-prover binary

Freyskeyd:
# Description

This PR is initiating the `aggkit-prover` binary, refactoring a bit the different crates to reuse code as much as possible.

## PR Checklist:


crates/aggkit-prover-config/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+
+pub mod shutdown;
+pub(crate) mod telemetry;
+
+pub(crate) const DEFAULT_IP: std::net::Ipv4Addr = std::net::Ipv4Addr::new(0, 0, 0, 0);
+
+/// The Aggkit Prover configuration.
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub struct ProverConfig {
```
atanmarko, line 22: I am confused here. Now we have duplicated `ProverConfig`, one for `agglayer-prover` and the other for `aggkit-prover`? I perfer that we start clean, and add features that we need.

atanmarko: Maybe we could keep common `ProverConfig` as it is in the `prover-config` crate, and add `AgglayerProverConfig` and `AggkitProverConfig` to include things there that differ with the common field `prover_config`

crates/aggkit-prover-config/tests/fixtures/validate_config/max_rpc_request_size.toml
```diff
@@ -0,0 +1,2 @@
+[rpc]
```
atanmarko, line 1: We should also keep to common test configurations on one place. Add only specified `aggkit-prover-config` and `agglayer-prover-config` to their respective folders.

crates/aggkit-prover/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+license.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+anyhow.workspace = true
+bincode.workspace = true
+futures.workspace = true
+# thiserror.workspace = true
```
atanmarko, line 14: Cleanup all, then add per need

crates/aggkit-prover-config/src/lib.rs
Freyskeyd: > Maybe we could keep common `ProverConfig` as it is in the `prover-config` crate, and add `AgglayerProverConfig` and `AggkitProverConfig` structures to include things there that differ with the common field `prover_config`

The `prover-*` should not link to any of `agglayer-` or `aggkit-`, it is a "framework" that can be used to compose further prover and let the community build their own if they want to. The configs of `agglayer-prover` and `aggkit-prover` will not have the same shape at all at the end.

atanmarko: Ok but still lets unify `agglayer-prover-config` and `aggkit-prover-config` in the common crate. Config customizations in `aggkit-prover` and `agglayer-prover`.

atanmarko (approved):
Ok, lets merge this to get us started, will refactor further as needed.

PR #23, opened 2025-02-04
refactor: removing pessimistic-proof deps from provers

Freyskeyd:
# Description

This PR is removing the `pessimistic-proof` and `agglayer-types` crate from the `provers`.
It resolves a circular dependency that has been pending since the move of agglayer-prover to this repo.

> [!NOTE]
> There is still the `agglayer-telemetry` crate that need to be decoupled.

## PR Checklist:


proto/agglayer/prover/v1/proof_generation.proto
```diff
@@ -4,16 +4,18 @@ package agglayer.prover.v1;
 
 // Service for generating pessimistic proof.
 service PessimisticProofService {
-  // Triggers a pessimistic-proof generation for a given batch header and initial state.
+  // Triggers a pessimistic-proof generation for a given batch header and
+  // initial state.
```
hadjiszs, line 8:
```suggestion
  // Triggers a pessimistic-proof generation.
```

hadjiszs (dismissed):

crates/agglayer-prover-types/src/error.rs
```diff
@@ -20,8 +19,8 @@ pub enum Error {
     #[error("Prover verification failed: {0}")]
     ProofVerificationFailed(#[from] ProofVerificationError),
 
-    #[error("Prover executor failed: {0}")]
-    ExecutorFailed(#[from] ProofError),
+    #[error("Prover executor failed")]
+    ExecutorFailed(Vec<u8>),
```
Ekleog-Polygon, line 23: Do we want that as a `Vec<u8>` or as a human-facing `String`?

crates/prover-executor/src/tests.rs
```diff
@@ -30,22 +45,9 @@ async fn executor_normal_behavior() {
...
-    let mut state = LocalNetworkStateData::default();
-    let certificate = Certificate::new_for_test(0.into(), 0);
-    let signer = certificate.get_signer();
-    let batch_header = state
-        .apply_certificate(
-            &certificate,
-            signer,
-            certificate.l1_info_root().unwrap().unwrap_or_default(),
-        )
-        .unwrap();
```
Ekleog-Polygon, line 43: I’m not sure I understand, why are we removing all the certificates from the tests here and in the rest of this file? AFAIU `SP1Stdin::new()` will not have any certificate inserted in the batch_header as things currently stand

crates/agglayer-prover/src/lib.rs
```diff
@@ -22,18 +16,17 @@ mod rpc;
 ///
 /// This function returns on fatal error or after graceful shutdown has
 /// completed.
-pub fn main(cfg: PathBuf, version: &str) -> Result<()> {
-    // Load the configuration file
-    let config: Arc<ProverConfig> = Arc::new(toml::from_str(&std::fs::read_to_string(cfg)?)?);
-
-    let global_cancellation_token = CancellationToken::new();
+pub fn main(cfg: PathBuf, version: &str, program: &'static [u8]) -> anyhow::Result<()> {
```
Ekleog-Polygon, line 19: I cannot find the place where this function would be called. Would we generate the agglayer-prover program from the regular `agglayer/agglayer` repo by just calling this function, or would we still somehow have to live with the dependency somewhere?

crates/prover-executor/src/tests.rs
Freyskeyd: Cause the `prover-executor` is agnostic of the program used with it

proto/agglayer/prover/v1/proof_generation.proto
```diff
@@ -4,16 +4,17 @@ package agglayer.prover.v1;
...
 message GenerateProofRequest {
-  // The initial state of the network used to generate the pessimistic proof.
-  bytes initial_state = 1;
-  // The batch header for which the proof is generated.
-  bytes batch_header = 2;
+  // Represents the type of stdin of the request
+  oneof stdin {
+    // STDIN using sp1 format
+    bytes sp1_stdin = 1;
+  }
```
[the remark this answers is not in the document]
Freyskeyd: We're the only one using it, and we've a 1:1 interaction, breaking are still allowed so, let's enjoy it ahah

crates/agglayer-prover-types/src/error.rs
Freyskeyd: The whole error management should be refactored I guess, let's merge this and iterate over it

PR #34, opened 2025-02-10
feat: add fields in aggchain proof

temaniarpit27:
Add new fields in Aggchain Proof


proto/aggkit/prover/v1/aggchain_proof_generation.proto
```diff
@@ -20,15 +20,22 @@ message GenerateAggchainProofRequest {
   bytes l1_info_tree_leaf_hash = 4;
   // L1 Info tree proof. ([32]hash)
   repeated bytes l1_info_tree_merkle_proof = 5;
+  // L1 info tree (hash => [32]hash)
+  map<string, Proof> ger_inclusion_proofs = 6;
```
Freyskeyd, line 24: The key should be a base64 encoded representation of the GER. Or, if we want to keep bytes, we need to fallback into the old `Entry` format.

```suggestion
  // Map of the GER with their inclusion proof. Note: the GER (string) is a base64 encoded string of the GER digest.
  map<string, Proof> ger_inclusion_proofs = 6;
```

proto/aggkit/prover/v1/aggchain_proof_generation.proto
```diff
@@ -16,19 +16,101 @@ message GenerateAggchainProofRequest {
...
+}
+
+// Encapsulates the information to uniquely identify a token on the origin
+// network.
+message TokenInfo {
+  // Network which the token originates from
+  uint32 origin_network = 1;
+  // The address of the token on the origin network
+  bytes origin_token_address = 2;
+}
```
Freyskeyd, line 116: All of this will be replaced by the proto definition in the agglayer repo. We can merge this as is, but it will be updated later :)

Freyskeyd (approved):

PR #54, opened 2025-02-17
test: adding tests to proposer & client

Freyskeyd:
# Description

This PR is adding test to proposer-client, proposer-service and update some struct.

## PR Checklist:


crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -62,7 +60,7 @@ pub struct AggchainProofBuilderRequest {
 #[derive(Clone, Debug)]
 pub struct AggchainProofBuilderResponse {
     /// Generated aggchain proof for the block range.
-    pub proof: AggchainProof,
+    pub proof: SP1Proof,
```
Ekleog-Polygon, line 63: I'm not sure I understand, is this on purpose? The `AggchainProof` type had seemed made just for this reason, but now we're replacing it with a simple `SP1Proof`?

crates/proposer-client/src/error.rs
```diff
@@ -2,6 +2,12 @@ use crate::ProofId;
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
+    #[error("Unable to create RPC client")]
+    UnableToCreateRPCClient(#[source] jsonrpsee::core::client::Error),
+
+    #[error("An error occurred while requesting an aggregated proof")]
+    AggProofRequestFailed(#[source] jsonrpsee::core::client::Error),
```
Ekleog-Polygon, line 9: ❤️

crates/proposer-client/src/tests/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        };
+
+        let response = service.request_agg_proof(request).await;
+
+        assert!(response.is_ok());
+        mock.assert_async().await;
+    }
+
+    #[test]
+    #[ignore = "to be implemented"]
```
Ekleog-Polygon, line 55: Nit: can we have the "todo" string somewhere, so it's easy to grep for it?

Ekleog-Polygon (commented):
Do I understand correctly that this will use JSON-RPC to handle all the RPC with the prover as well as the proposer? Why not use gRPC, for this new server-to-server piece of code?

crates/proposer-client/src/tests/mod.rs
Freyskeyd: I'm expecting those tests to be completed ASAP
I can put a todo for convenience

Freyskeyd:
> Do I understand correctly that this will use JSON-RPC to handle all the RPC with the prover as well as the proposer? Why not use gRPC, for this new server-to-server piece of code?

`proposer-client` is an existing jsonrpc go implementation.
If they provide the same capabilities with gRPC I'll be happy to switch, unfortunately it's not the case.

crates/aggchain-proof-builder/src/lib.rs
atanmarko: @Freyskeyd Are we sure we don't need additional metadata to put in `AgghchainProof`?

crates/proposer-client/src/rpc.rs
```diff
@@ -81,6 +92,7 @@ impl From<AggSpanProofProposerRequest> for ProposerRequest {
 /// Response for the external proposer `request_span_proof` call
 #[derive(Serialize, Deserialize, Debug)]
 pub struct AggSpanProofProposerResponse {
+    #[serde(rename = "proof_request_id")]
```
atanmarko, line 95: Why is this snake case and `AggSpanProofProposerRequest` members are camel case?

crates/prover-alloy/src/lib.rs
```diff
@@ -31,6 +30,15 @@ type AlloyFillProvider = FillProvider<
     Ethereum,
 >;
 
+#[async_trait]
+#[cfg_attr(feature = "testutils", mockall::automock)]
+pub trait Provider {
```
atanmarko, line 35: Just not sure about the trait name - could be confused with actual `alloy::providers::Provider`. Maybe `RpcProvider`?

crates/proposer-client/src/rpc.rs
Freyskeyd: The API is made that way

crates/prover-alloy/src/lib.rs
Freyskeyd: The goal is to not expose `alloy::providers::Provider` at some point, maybe we should rename this crate to something else :)

atanmarko: @Freyskeyd I know we will not expose `alloy::providers::Provider`, we have this crate to hide it behind. Just using the same trait name `Provider` could be confusing for somebody looking at the code where it is used.

atanmarko (approved):

PR #62, opened 2025-02-21
feat: adding customchaindata builder service

Freyskeyd:
# Description

This PR is adding the `customchaindata` service.

## PR Checklist:


crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -69,6 +69,8 @@ pub struct AggchainProofBuilderResponse {
     pub start_block: u64,
     /// Last block included in the aggchain proof.
     pub end_block: u64,
+    /// Output root
+    pub output_root: Hash,
```
Ekleog-Polygon, line 73: We have many roots in our systems, can we specify which root this is about exactly? AFAIU this would be `output_rollup_root`

crates/aggchain-proof-core/Cargo.toml
```diff
@@ -30,5 +30,6 @@ alloy.workspace = true
 dotenvy.workspace = true
 serde_json.workspace = true
 sp1-cc-host-executor = { git = "https://github.com/succinctlabs/sp1-contract-call/", rev = "c9bdace" }
+sp1-sdk = { workspace = true }
```
Ekleog-Polygon, line 33:
```suggestion
sp1-sdk.workspace = true
```

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+const AGGCHAIN_VKEY_SELECTOR: [u8; 4] =
+    calculate_aggchain_vkey_selector(AGGCHAIN_PROOF_PROGRAM_VERSION, AGGCHAIN_TYPE);
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {}
+
+#[derive(Debug)]
+pub(crate) struct CustomChainDataBuilderRequest {
+    pub(crate) l2_block_number: u64,
+    pub(crate) output_root: [u8; 32],
```
Ekleog-Polygon, line 21:
```suggestion
    pub(crate) output_root: Digest,
```

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+#[derive(Debug, PartialEq)]
+pub(crate) struct CustomChainDataBuilderResponse {
+    pub custom_chain_data: [u8; 66],
+}
+
+pub(crate) struct CustomChainDataBuilderService<L1Rpc> {
+    l1_rpc: Arc<L1Rpc>,
+    network_id: u32,
```
Ekleog-Polygon, line 31:
```suggestion
    network_id: NetworkId,
```

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+    vkey: Arc<[u32; 8]>,
```
Ekleog-Polygon, line 32: I… think we already have a type for a VKey somewhere?

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+impl Clone for CustomChainDataBuilderService<L1Rpc> {
```
Ekleog-Polygon, line 35: Why not just derive Clone? (If it's for automatically-defined bounds reasons, `educe` can ignore the bounds)

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    fn call(&mut self, req: CustomChainDataBuilderRequest) -> Self::Future {
+        let rpc = self.l1_rpc.clone();
+
+        let vkey = self.vkey.clone();
+        async move {
+            let registered_vkey = rpc.resolve_vkey(AGGCHAIN_VKEY_SELECTOR).await?;
+            if *vkey != registered_vkey {
+                error!(
```
Ekleog-Polygon, line 74: Does this deserve bubbling up an error? After all, we're bubbling up an error if L1 is unreachable already

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+    async fn resolve_vkey(&self, aggchain_vkey_selector: [u8; 4]) -> Result<[u32; 8], Error>;
+}
+
+pub struct L1Rpc {}
+
+#[async_trait::async_trait]
+impl AggchainVkeyResolver for L1Rpc {
+    async fn resolve_vkey(&self, _aggchain_vkey_selector: [u8; 4]) -> Result<[u32; 8], Error> {
+        // Implementation for resolving vkey from L1 RPC
+        unimplemented!()
```
Ekleog-Polygon, line 104:
```suggestion
        todo!()
```

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+    selector: u16,
+    output_root: [u8; 32],
+    l2_block_number: u64,
+) -> [u8; 66] {
+    let mut custom_chain_data = [0u8; 66];
+
+    // Convert values to bytes in big-endian order
+    custom_chain_data[0..2].copy_from_slice(&selector.to_be_bytes());
+    custom_chain_data[2..34].copy_from_slice(&output_root);
+    custom_chain_data[58..66].copy_from_slice(&l2_block_number.to_be_bytes());
```
Ekleog-Polygon, line 118: There's a hole between 34 and 58, is it on purpose?

crates/aggkit-prover-types/src/compat/mod.rs
```diff
@@ -0,0 +1,125 @@
+use prost::bytes::Bytes;
+
+use crate::agglayer::protocol::types::v1::{FixedBytes20, FixedBytes32, FixedBytes65};
+
+impl From<[u8; 32]> for FixedBytes32 {
```
Ekleog-Polygon, line 5: This file is also duplicated from agglayer… makes me some more sad 😢

Ekleog-Polygon (changes requested):
Overall looks good to me! Unfortunately, I wrote my comments as I was reading, so the nits are mixed in with the few important ones.

But there are a few ones I think are important to address:
* Removal of the unneeded build.rs
* Hardcoded network id and root that are probably just WIP but not marked with a comment yet
* Using strong types if possible
* `custom_chain_data` hole

crates/aggchain-proof-service/src/service.rs
```diff
@@ -90,6 +100,17 @@ impl AggchainProofService {
         .map_err(Error::AlloyProviderInitializationFailed)?;
         let l1_rpc_client = Arc::new(client);
 
+        // TODO: Remplace this DUMMY vkey by the reel one
+        let vkey = [0u32; 8];
+        let chain_l1_rpc = L1Rpc {};
```
atanmarko, line 105: You have this in the `config.proposer_service.l1_rpc_endpoint`. Probably best to make the config for the `CustomChainDataBuilderService`.

crates/aggchain-proof-service/src/service.rs
```diff
@@ -90,6 +100,17 @@ impl AggchainProofService {
...
+        let customchaindata_builder = tower::ServiceBuilder::new()
+            .service(CustomChainDataBuilderService::new(
+                Arc::new(chain_l1_rpc),
+                1,
```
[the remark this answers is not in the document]
atanmarko: You already have `network_id` in the `config.aggchain_proof_builder` in this function.

crates/aggchain-proof-service/src/service.rs
```diff
@@ -137,38 +159,50 @@ impl tower::Service<AggchainProofServiceRequest> for AggchainProofService {
...
+
+            let aggchain_proof_response = proof_builder
+                .call(aggchain_proof_builder_request)
+                .await
+                .map_err(Error::AggchainProofBuilderServiceError)?;
+
+            let custom_chain_data = customchaindata_builder
+                .call(CustomChainDataBuilderRequest {
+                    l2_block_number: agg_span_proof_response.end_block,
+                    output_root: [1u8; 32],
```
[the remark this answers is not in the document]
atanmarko: You can use `aggchain_proof_response.output_root`.

crates/aggchain-proof-service/src/service.rs
```diff
@@ -137,38 +159,50 @@ impl tower::Service<AggchainProofServiceRequest> for AggchainProofService {
...
+                })
+                .await
+                .map_err(Error::CustomChainDataBuilderError)?;
+
+            Ok(AggchainProofServiceResponse {
+                proof: aggchain_proof_response.proof,
+                start_block: agg_span_proof_response.start_block,
+                end_block: agg_span_proof_response.end_block,
+                local_exit_root_hash: Default::default(),
```
atanmarko, line 202: Why `default`? Add it to the `AggchainProofBuilderResponse` and use it here.

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
+use std::{
```
atanmarko, line 1: Move this service to the separate crate, e.g. `aggchain-custom-chain-data-builder` . 
Add Error and Config types for it. 
It should be consistent with the other services.

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+
+#[derive(Debug)]
+pub(crate) struct CustomChainDataBuilderRequest {
+    pub(crate) l2_block_number: u64,
+    pub(crate) output_root: [u8; 32],
+}
+
+#[derive(Debug, PartialEq)]
+pub(crate) struct CustomChainDataBuilderResponse {
+    pub custom_chain_data: [u8; 66],
```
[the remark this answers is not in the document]
atanmarko: Even better the dedicated type for `CustomChainData` here or in the `aggkit-prover-types`.

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
atanmarko: Yes, this is straightforward, no reason to implement it manually.

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: CustomChainDataBuilderRequest) -> Self::Future {
+        let rpc = self.l1_rpc.clone();
+
+        let vkey = self.vkey.clone();
```
iljakuklic, line 70: Seems vkey does not need to be cloned here. Taking a reference suffices for the comparison below.

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+            })
+        }
+        .boxed()
+    }
+}
+
+#[async_trait::async_trait]
+#[cfg_attr(test, mockall::automock)]
+trait AggchainVkeyResolver {
+    async fn resolve_vkey(&self, aggchain_vkey_selector: [u8; 4]) -> Result<[u32; 8], Error>;
```
iljakuklic, line 95: A custom type for `VKeySelector` would be nice.

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+        unimplemented!()
+    }
+}
+
+pub fn calculate_custom_chain_data(
+    selector: u16,
+    output_root: [u8; 32],
+    l2_block_number: u64,
+) -> [u8; 66] {
+    let mut custom_chain_data = [0u8; 66];
```
iljakuklic, line 113: Is the custom chain data always in this format, i.e. selector, output root, (gap), l2 block? If so, could it just be a type with these fields that serializes / deserializes to the appropriate byte sequence?

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
```diff
@@ -0,0 +1,281 @@
...
+
+    custom_chain_data
+}
+
+pub const fn calculate_aggchain_vkey_selector(program: u16, aggchain_type: u16) -> [u8; 4] {
+    (((program as u32) << 16) | aggchain_type as u32).to_be_bytes()
+}
+
+#[cfg(test)]
+mod tests {
```
iljakuklic, line 128: Nit: move tests into a separate file?

iljakuklic (commented):
Agreed with most of the points already risen by others. Some extra nits below.

atanmarko:
Note: maybe this PR could be split in two, one for `buf` changes and other for custom chain data builder?

crates/aggchain-proof-service/build.rs
```diff
@@ -0,0 +1,37 @@
...
+        .expect("Unable to extract version");
+
+    let major_version = version.major.to_string();
+
+    let out_dir = "./src/context/";
+    let dest_path = Path::new(&out_dir).join("aggchain_proof_program_version.rs");
+    fs::write(
+        &dest_path,
+        format!(
+            "pub const AGGCHAIN_PROOF_PROGRAM_VERSION: u16 = {};\n",
```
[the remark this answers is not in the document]
Freyskeyd: `CARGO_PKG_VERSION` is returning the version of the current package if I'm not wrong, here, the goal is to extract the version of the `program`

Freyskeyd:
Putting this PR back in draft to clean up, I think I messed up something, will re-ask for review when ready !

crates/aggchain-proof-service/build.rs
```diff
@@ -0,0 +1,37 @@
...
+    let version: Version = parsed_toml
+        .get("package")
+        .and_then(|pkg| pkg.get("version"))
+        .and_then(|v| {
+            v.as_str()
+                .map(Version::parse)
+                .transpose()
+                .expect("Unable to extract version")
+        })
+        .expect("Unable to extract version");
```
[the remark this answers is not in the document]
Freyskeyd: I'm not used to `vergen`, but readying the `env` will not produce a `const u16` but a `str` that need to be parsed.

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
Freyskeyd: We have `L1Rpc` as generic and `#[derive(Clone)]` expects that all generic implement `Clone`.
In our case, we don't want to have `L1Rpc` to be `Clone`, we just want to clone the `Arc` but the compiler doesn't pick that.

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
Freyskeyd: Postpone to another PR

crates/aggchain-proof-service/src/service.rs
Freyskeyd: Because it was already `Default::default` and there is no `local_exit_root_hash` on the `AggchainProofBuilderResponse` If I'm not wrong

crates/aggchain-proof-service/src/service/customchaindata_builder.rs
Freyskeyd: I don't think we can because of lifetime

crates/aggchain-proof-core/build.rs
```diff
@@ -0,0 +1,37 @@
...
+    fs::write(
+        &dest_path,
+        format!(
+            "pub const AGGCHAIN_PROOF_PROGRAM_VERSION: u16 = {};\n",
+            major_version
+        ),
+    )
+    .expect("Failed to write aggchain_proof_program_version.rs");
+
+    println!("cargo:rerun-if-changed=Cargo.toml");
```
iljakuklic, line 36: Does this watch changes for the correct `Cargo.toml`? I'd expect it to refer to `../aggchain-proof-program/Cargo.toml"`.

Also, this line will not be emitted in any of the steps above fail. Is that correct? Maybe it should be higher up somewhere around where the cargo file is loaded.

crates/aggchain-proof-core/build.rs
```diff
@@ -0,0 +1,37 @@
...
+    let out_dir = "./src/";
+    let dest_path = Path::new(&out_dir).join("version.rs");
+    fs::write(
+        &dest_path,
+        format!(
+            "pub const AGGCHAIN_PROOF_PROGRAM_VERSION: u16 = {};\n",
+            major_version
+        ),
+    )
+    .expect("Failed to write aggchain_proof_program_version.rs");
```
iljakuklic, line 34: I think this should go into a build output directory rather than overwriting files in the source. There's [an example](https://doc.rust-lang.org/cargo/reference/build-script-examples.html#code-generation) of this in cargo docs.

Also, it is a good practice to emit a comment into the file saying it's been auto-generated and should not be modified manually.

crates/aggchain-proof-service/src/custom_chain_data/mod.rs
```diff
@@ -0,0 +1,41 @@
+use aggchain_proof_core::{Digest, AGGCHAIN_PROOF_PROGRAM_VERSION, AGGCHAIN_TYPE};
+use alloy_primitives::U256;
+use bincode::Options;
+use serde::{Deserialize, Serialize};
+
+#[cfg(test)]
+mod tests;
+
+type VKeySelector = [u8; 4];
```
iljakuklic, line 9: Nit: I'd prefer this to be a strong type rather than alias. Would allow you to replace the freestanding function `compute_aggchain_vkey_selector` with something like `VKeySelector::new`.

Makefile.contracts.toml
```diff
@@ -11,7 +11,7 @@ dependencies = [
     { name = "contracts-download-PolygonRollupManager" },
     { name = "contracts-download-PolygonZkEVM" },
     { name = "contracts-download-PolygonZkEVMGlobalExitRootV2" },
-    { name = "contracts-download-AggchainFEP" },
+    { name = "contracts-download-AggchainFEP" }
```
Ekleog-Polygon, line 14: Can we undo this change? Trailing commas usually good for whenever we need to change the items, and toml is supposed to allow trailing commas

crates/aggchain-proof-service/build.rs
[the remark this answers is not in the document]
Ekleog-Polygon: About non-constness, we can still do that without issue: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c56e3bf6a85f250c595b5f5e720e54f8

Unfortunately, I did miss the fact that we're extracting the version of the `program`. Unfortunately there are no artifact dependencies in cargo yet, so no good way to express this dependency 😢 

(marking as unresolved just so this doesn't get lost in the review comments, feel free to mark as resolved after reading!)

crates/aggchain-proof-service/src/custom_chain_data/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+
+// Making this unused as it will be used in another iteration
```
Ekleog-Polygon, line 11:
```suggestion
// TODO: Making this unused as it will be used in another iteration
```

crates/aggchain-proof-service/src/custom_chain_data/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+    compute_aggchain_vkey_selector(AGGCHAIN_PROOF_PROGRAM_VERSION, AGGCHAIN_TYPE);
+
+#[derive(Serialize, Deserialize)]
+pub(crate) struct CustomChainData {
+    selector: u16,
+    output_root: Digest,
+    l2_block_number: [u8; U256::BYTES],
+}
+
+// Making this unused as it will be used in another iteration
```
Ekleog-Polygon, line 23:
```suggestion
// TODO: Making this unused as it will be used in another iteration
```

crates/aggchain-proof-service/src/service.rs
```diff
@@ -137,38 +150,47 @@ impl tower::Service<AggchainProofServiceRequest> for AggchainProofService {
...
+                aggchain_proof_response.output_root,
+                agg_span_proof_response.end_block,
+            )
+            .map_err(Error::UnableToSerializeCustomChainData)?;
+
+            Ok(AggchainProofServiceResponse {
+                proof: aggchain_proof_response.proof,
+                start_block: agg_span_proof_response.start_block,
+                end_block: agg_span_proof_response.end_block,
+                local_exit_root_hash: Default::default(),
```
Ekleog-Polygon, line 190:
```suggestion
                local_exit_root_hash: Default::default(), // TODO
```

Ekleog-Polygon (commented):
LGTM! Besides the changes requested by @iljakuklic, I just have a few TODO annotations to suggest, so that we don't forget places we know we need to come back to :)

crates/aggchain-proof-service/build.rs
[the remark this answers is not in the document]
Freyskeyd: Yeah I endup with the same outcome on this subject.
I scratch my head to find a way to use pre-existing "workspace" env var but there is a missing piece for such kind of mechanism.
I'll let this as is, it is good enough for a first iteration and we'll improve that if needed later.

Ekleog-Polygon (dismissed):
LGTM 🎉

crates/aggchain-proof-service/src/custom_chain_data/mod.rs
```diff
@@ -6,12 +6,26 @@ use serde::{Deserialize, Serialize};
 #[cfg(test)]
 mod tests;
 
-type VKeySelector = [u8; 4];
+#[derive(Debug)]
+// TODO: Making this unused as it will be used in another iteration
+#[allow(unused)]
+pub struct VKeySelector([u8; 4]);
```
iljakuklic, line 12: Does `#[expect()]` work here? That way we'll get heads up from the compiler if it becomes used but it does not always work with all possible build options (test vs non-test, various features enabled, etc).

```suggestion
#[expect(unused)]
pub struct VKeySelector([u8; 4]);
```

iljakuklic (dismissed):

crates/aggchain-proof-service/src/custom_chain_data/mod.rs
Freyskeyd: After trying, it doesn't work well I think.
I will merge it as is

PR #67, opened 2025-02-25
feat: uses protocol buf types

Freyskeyd:
# Description

This PR refactors the protobuf files to use agglayer's protocol types.

## PR Checklist:


crates/aggchain-proof-service/src/service.rs
```diff
@@ -130,38 +130,33 @@ impl tower::Service<AggchainProofServiceRequest> for AggchainProofService {
...
+            let agg_span_proof_response = proposer_service.call(proposer_request).await?;
+            let aggchain_proof_builder_request =
+                aggchain_proof_builder::AggchainProofBuilderRequest {
+                    agg_span_proof: agg_span_proof_response.agg_span_proof,
+                    start_block: agg_span_proof_response.start_block,
+                    end_block: agg_span_proof_response.end_block,
+                    l1_info_tree_merkle_proof: req.l1_info_tree_merkle_proof,
+                    l1_info_tree_leaf: req.l1_info_tree_leaf,
+                    l1_info_tree_root_hash: req.l1_info_tree_root_hash,
+                    ger_inclusion_proofs: req.ger_inclusion_proofs,
```
temaniarpit27, line 146: @Freyskeyd not sure from where this is coming.

crates/aggkit-prover-types/src/lib.rs
```diff
@@ -3,12 +3,14 @@ use bincode::{
...
 };
 
-pub const FILE_DESCRIPTOR_SET: &[u8] = include_bytes!("generated/aggkit.prover.bin");
+pub mod compat;
 
-#[path = "generated/aggkit.prover.v1.rs"]
 #[rustfmt::skip]
 #[allow(warnings)]
-pub mod v1;
+mod generated;
```
atanmarko, line 10: `generated` is a bit too generic :). I would argue that `protobuf` or `protoc` or similar would be a better choice.

atanmarko (approved):

crates/aggkit-prover-types/src/lib.rs
Freyskeyd: That's the generic name that we're using in every crate that has generated protobuf file. And this folder isn't exposed publically.

crates/aggchain-proof-service/src/service.rs
temaniarpit27: @Freyskeyd so we will need to change here right?

PR #71, opened 2025-02-26
feat: aggchain contracts client

### LINKED ISSUE #46: Retrieve the `l2_pre_root` and `claim_root` from the l2

### LINKED ISSUE #47: Retrieve the `rollup_config_hash` from the `AggchainFEP.sol`

atanmarko:
# Description

Introduces generic `AggchainContractsClient` used by the `aggchain-proof-builder` to retrieve public inputs from the chain smart contracts.

In this PR implemented are retrieval of the `l2_pre_root` and the `claim_root` and their pre-image data. It also retrieves rollup config hash from the `AggchainFep` l1 contract.

Fixes #46
Fixes #47

Note:
Additional client for the l2 rollup node is also added.

## PR Checklist:

### DIFF
--- a/crates/aggchain-proof-contracts/src/config.rs
+++ b/crates/aggchain-proof-contracts/src/config.rs
@@ -1,0 +1,46 @@
+use alloy::primitives::{address, Address};
+use prover_utils::from_env_or_default;
+use serde::{Deserialize, Serialize};
+
+/// Address of the `GlobalExitRootManagerL2SovereignChain.sol` contract
+/// on the L2 chain is always fixed.
+const GLOBAL_EXIT_ROOT_MANAGER_L2_SOVEREIGN_CHAIN_ADDRESS: Address =
+    address!("0xa40D5f56745a118D0906a34E69aeC8C0Db1cB8fA");
+
+/// Address of the `PolygonRollupManager.sol` contract
+/// on the L1 chain.
+const POLYGON_ROLLUP_MANAGER: Address = address!("0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e");
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+#[serde(rename_all = "kebab-case")]
+pub struct AggchainProofContractsConfig {
+    #[serde(alias = "PolygonRollupManagerContract")]
+    pub polygon_rollup_manager_contract: Address,
+    #[serde(alias = "GlobalExitRootManagerL2SovereignChainContract")]
+    pub global_exit_root_manager_v2_sovereign_chain: Address,
+}
+
+impl Default for AggchainProofContractsConfig {
+    fn default() -> Self {
+        Self {
+            polygon_rollup_manager_contract: default_polygon_rollup_manager(),
+            global_exit_root_manager_v2_sovereign_chain:
+                default_global_exit_root_manager_v2_sovereign_chain(),
+        }
+    }
+}
+
+pub(crate) fn default_output_at_block_endpoint() -> String {
+    from_env_or_default(
+        "L2_OUTPUT_AT_BLOCK_ENDPOINT",
+        "l2_outputAtBlock".to_string(),
+    )
+}
+
+fn default_polygon_rollup_manager() -> Address {
+    POLYGON_ROLLUP_MANAGER
+}
+
+fn default_global_exit_root_manager_v2_sovereign_chain() -> Address {
+    GLOBAL_EXIT_ROOT_MANAGER_L2_SOVEREIGN_CHAIN_ADDRESS
+}
--- a/crates/aggchain-proof-contracts/src/config.rs
+++ b/crates/aggchain-proof-contracts/src/config.rs
@@ -14,17 +14,19 @@ const POLYGON_ROLLUP_MANAGER: Address = address!("0xB7f8BC63BbcaD18155201308C8f3
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
 #[serde(rename_all = "kebab-case")]
 pub struct AggchainProofContractsConfig {
-    #[serde(alias = "PolygonRollupManagerContract")]
+    /// Address of the L1 PolygonRollupManager.sol contract
+    #[serde(default = "default_polygon_rollup_manager")]
     pub polygon_rollup_manager_contract: Address,
-    #[serde(alias = "GlobalExitRootManagerL2SovereignChainContract")]
-    pub global_exit_root_manager_v2_sovereign_chain: Address,
+    /// Address of the L2 GlobalExitRootManagerL2SovereignChain.sol contract
+    #[serde(default = "default_global_exit_root_manager_v2_sovereign_chain")]
+    pub global_exit_root_manager_v2_sovereign_chain_contract: Address,
 }
 
 impl Default for AggchainProofContractsConfig {
     fn default() -> Self {
         Self {
             polygon_rollup_manager_contract: default_polygon_rollup_manager(),
-            global_exit_root_manager_v2_sovereign_chain:
+            global_exit_root_manager_v2_sovereign_chain_contract:
                 default_global_exit_root_manager_v2_sovereign_chain(),
         }
     }
--- a/crates/aggchain-proof-contracts/src/config.rs
+++ b/crates/aggchain-proof-contracts/src/config.rs
@@ -16,17 +16,17 @@ const POLYGON_ROLLUP_MANAGER: Address = address!("0xB7f8BC63BbcaD18155201308C8f3
 pub struct AggchainProofContractsConfig {
     /// Address of the L1 PolygonRollupManager.sol contract
     #[serde(default = "default_polygon_rollup_manager")]
-    pub polygon_rollup_manager_contract: Address,
+    pub polygon_rollup_manager: Address,
     /// Address of the L2 GlobalExitRootManagerL2SovereignChain.sol contract
     #[serde(default = "default_global_exit_root_manager_v2_sovereign_chain")]
-    pub global_exit_root_manager_v2_sovereign_chain_contract: Address,
+    pub global_exit_root_manager_v2_sovereign_chain: Address,
 }
 
 impl Default for AggchainProofContractsConfig {
     fn default() -> Self {
         Self {
-            polygon_rollup_manager_contract: default_polygon_rollup_manager(),
-            global_exit_root_manager_v2_sovereign_chain_contract:
+            polygon_rollup_manager: default_polygon_rollup_manager(),
+            global_exit_root_manager_v2_sovereign_chain:
                 default_global_exit_root_manager_v2_sovereign_chain(),
         }
     }
--- a/crates/aggchain-proof-contracts/src/config.rs
+++ b/crates/aggchain-proof-contracts/src/config.rs
@@ -1,6 +1,7 @@
 use alloy::primitives::{address, Address};
 use prover_utils::from_env_or_default;
 use serde::{Deserialize, Serialize};
+use url::Url;
 
 /// Address of the `GlobalExitRootManagerL2SovereignChain.sol` contract
 /// on the L2 chain is always fixed.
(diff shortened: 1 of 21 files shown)


crates/aggchain-proof-contracts/src/lib.rs
```diff
@@ -103,38 +196,89 @@ impl AggchainProofContractsClient {
...
+                .map_err(Error::AsyncEngineSetupError)?
+                .block_on(async move {
+                    // We need to retrieve AggchainFep contract address from the
+                    // polygon rollup manager.
+                    let rollup_data = polygon_rollup_manager
+                        .rollupIDToRollupData(network_id)
+                        .call()
+                        .await?;
+                    Ok(rollup_data.rollupData.rollupContract)
+                })
```
Ekleog-Polygon, line 223: This feels wrong. Can we just make the function be async from the beginning? It could use lots of `spawn_blocking`, but that way would feel better to me than spawning a whole new tokio runtime just to call one async function

Ekleog-Polygon (commented):
Overall LGTM! I'll go through another review round once the comments are fixed/answered :)

crates/aggchain-proof-contracts/src/config.rs
```diff
@@ -0,0 +1,46 @@
...
+    )
+}
+
+fn default_polygon_rollup_manager() -> Address {
+    POLYGON_ROLLUP_MANAGER
+}
+
+fn default_global_exit_root_manager_v2_sovereign_chain() -> Address {
+    GLOBAL_EXIT_ROOT_MANAGER_L2_SOVEREIGN_CHAIN_ADDRESS
+}
```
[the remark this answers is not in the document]
atanmarko: For two reasons:
- It is quite likely that we may need to change this function and have some logic that would learn default addresses (e.g. parsing env or similar), and from discussion we had, they are fixed by nature, so don't deserve a configuration parameter.  And it is always easier to change function initializer.
- I want to be consistent, we have used functions initializers across the project

So I prefer to keep them like this, no real harm there 😉

crates/aggchain-proof-contracts/src/lib.rs
```diff
@@ -1,95 +1,188 @@
...
+            .call_raw()
+            .block(block_number.into())
+            .await
+            .map_err(Error::LocalExitRootError)?;
+
+        let result = self
+            .polygon_zkevm_bridge_v2
+            .getRoot()
+            .decode_output(response, true)
+            .map_err(Error::LocalExitRootError)?;
```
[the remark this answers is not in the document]
atanmarko: @Ekleog-Polygon I am afraid of exponential bloat of our error types if we take this direction. Details of what failed are already included in the source `alloy::contract::Error` so no information is lost or unclear.

crates/aggchain-proof-contracts/src/lib.rs
```diff
@@ -1,95 +1,188 @@
...
-                l2_rpc_endpoint,
-                prover_alloy::DEFAULT_HTTP_RPC_NODE_INITIAL_BACKOFF_MS,
-                prover_alloy::DEFAULT_HTTP_RPC_NODE_BACKOFF_MAX_RETRIES,
-            )
-            .map_err(Error::ProviderInitializationError)?,
+        let chain_config_hash_response = self
+            .aggchain_fep
+            .chainConfigHash()
+            .decode_output(response, true)
+            .map_err(Error::RollupConfigHashError)?;
```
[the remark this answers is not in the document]
atanmarko: You are totally right, I'll optimize this.

crates/aggchain-proof-contracts/src/lib.rs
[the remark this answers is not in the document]
atanmarko: EDIT: Optimized, so problem with connection or decoding will be communicated from the alloy `call()`

crates/aggchain-proof-contracts/src/lib.rs
```diff
@@ -1,95 +1,188 @@
...
+            prover_alloy::DEFAULT_HTTP_RPC_NODE_INITIAL_BACKOFF_MS,
+            prover_alloy::DEFAULT_HTTP_RPC_NODE_BACKOFF_MAX_RETRIES,
+        )
+        .map_err(Error::ProviderInitializationError)?;
+
+        let l2_cl_client = Arc::new(
+            HttpClient::builder()
+                .build(l2_cl_rpc_endpoint)
+                .map_err(Error::RollupNodeInitError)?,
         );
```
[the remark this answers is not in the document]
atanmarko: For the reason that CL client is custom JSON RPC api, and EL client is custom ETH smart contracts getter calls. In theory you could use `HttpClient` to call smart contract functions, but with the enormous amount of boilerplate that `alloy` hides from you.

crates/aggchain-proof-contracts/src/lib.rs
atanmarko: Whole setup of the stack and services, GRPC service,  `aggchain-proof-service`, `aggbhain-proof-builder` is not async. And it would be wrong to make it async just because of this. On the other side, to properly setup `agghcain-proof-contracts` we simply need to query some smart contracts for other smart contract addresses.

atanmarko: To mention the fact, we make the simple current thread limited runtime here, so it shouldn't be a big overhead.

Ekleog-Polygon (approved):
LGTM! I'm still a bit sad about the fact that we have async-in-sync-in-async, but we can live with it 🤷

atanmarko:
> LGTM! I'm still a bit sad about the fact that we have async-in-sync-in-async, but we can live with it 🤷

@Ekleog-Polygon What do you mean? We have async in sync on one place, and async in sync on another.

Ekleog-Polygon (dismissed):
I think it's what I mean? Basically we have an overall async runtime that calls a sync function with `spawn_blocking`, which somewhere down the call stack re-builds an async runtime in order to call two functions.

And actually, when I said "with spawn_blocking" I noticed that… maybe this is actually wrong?

I tried going up the sync chain, and for example in `testing_rpc_failure` we're using that function as though it were non-blocking… which it probably was before this addition? This is the only obvious end-user I could find, as I'm not very familiar with how this code actually gets called.

Overall I'd really feel more comfortable propagating the `async` up the call chain so that it's obvious the function can block.

I'm marking as "request changes" for now because of that question. If you can confirm that you're sure any call to this either comes from a sync `main` or from a `spawn_blocking`, I'll be happy to remove that request!

atanmarko:
@Ekleog-Polygon I am not following. `AggchainContractsRpcClient::new()` function where we instantiate current thread tokio runtime is called from the `AggchainProofBuilder::new()` which is called from the  `AggchainProofService::new()` `GrpcService::new` which is called from the `aggkit-prover main()` function which is a sync program, not async.

Ekleog-Polygon (commented):
Got it, thanks! So the only problematic place would be the test I had been pointing out, thank you for the clarification :) (Obviously the only place I found had to be the only one that was wrong 😆 )

atanmarko:
@Ekleog-Polygon New issue added for those tests https://github.com/agglayer/provers/issues/74

Ekleog-Polygon (approved):
LGTM!

crates/aggchain-proof-contracts/src/config.rs
```diff
@@ -0,0 +1,46 @@
...
+/// Address of the `PolygonRollupManager.sol` contract
+/// on the L1 chain.
+const POLYGON_ROLLUP_MANAGER: Address = address!("0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e");
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone)]
+#[serde(rename_all = "kebab-case")]
+pub struct AggchainProofContractsConfig {
+    #[serde(alias = "PolygonRollupManagerContract")]
+    pub polygon_rollup_manager_contract: Address,
+    #[serde(alias = "GlobalExitRootManagerL2SovereignChainContract")]
```
Freyskeyd, line 19: Why is there an `alias` ?
Should we define a `#[serde(default = "default_global_exit_root_manager_v2_sovereign_chain")]` ?

crates/aggchain-proof-contracts/src/lib.rs
```diff
@@ -1,140 +1,270 @@
...
+    fn new(
+        l1_rpc_endpoint: &Url,
+        l2_el_rpc_endpoint: &Url,
+        l2_cl_rpc_endpoint: &Url,
+        network_id: u32,
+        contracts: &AggchainProofContractsConfig,
+    ) -> Result<Self, Error>
+    where
+        Self: std::marker::Sized;
+}
```
Freyskeyd, line 39: This method should be part of `AggchainContractsRpcClient` not this `trait`

crates/aggchain-proof-contracts/src/lib.rs
```diff
@@ -1,140 +1,270 @@
...
+        l1_rpc_endpoint: &Url,
+        l2_el_rpc_endpoint: &Url,
+        l2_cl_rpc_endpoint: &Url,
+        network_id: u32,
+        contracts: &AggchainProofContractsConfig,
+    ) -> Result<Self, crate::Error> {
+        let async_runtime = tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .map_err(Error::AsyncEngineSetupError)?;
```
Freyskeyd, line 149: You don't need to create a new runtime I guess, it'll be run in an existing runtime

crates/aggchain-proof-contracts/src/lib.rs
```diff
@@ -1,140 +1,270 @@
...
-            tokio::runtime::Builder::new_current_thread()
-                .enable_all()
-                .build()
-                .map_err(Error::AsyncEngineSetupError)?
-                .block_on(async move {
-                    // We need to retrieve PolygonZkEVMBridgeV2 contract address from the
-                    // global exit root manager contract.
-                    global_exit_root_manager_l2.bridgeAddress().call().await
-                })
+            async_runtime.block_on(async move {
```
Freyskeyd, line 180: The block_on is on purpose or could we have this method async?

crates/aggchain-proof-contracts/src/config.rs
atanmarko: So I am not sure about that, maybe contracts should be referred with their original name to prevent confusion?

Freyskeyd: The agglayer-node approach is outdated and we had this alias for backward compatibility with the go-version.

atanmarko: Ok, lets switch then to `kebab-case`.

crates/aggchain-proof-contracts/src/lib.rs
Freyskeyd: I don't mind having all of those methods async

atanmarko: Ok, lets switch to async then.

crates/aggchain-proof-contracts/src/contracts.rs
```diff
@@ -0,0 +1,73 @@
...
+    AggchainFep::AggchainFepInstance<(), RpcProvider, Ethereum>;
+
+#[async_trait::async_trait]
+pub trait L2LocalExitRootFetcher {
+    async fn get_l2_local_exit_root(&self, block_number: u64) -> Result<Digest, Error>;
+}
+
+#[async_trait::async_trait]
+pub trait L2OutputAtBlockFetcher {
+    fn parse_l2_output_root(json: serde_json::Value) -> Result<L2OutputAtBlock, Error>;
```
Freyskeyd, line 56: This function is related to the internal of the implementation, it shouldn't be on the trait I guess

crates/aggchain-proof-contracts/src/lib.rs
```diff
@@ -1,140 +1,247 @@
...
+}
+
+impl AggchainContractsRpcClient<AlloyFillProvider> {
+    pub async fn new(
+        l1_rpc_endpoint: &Url,
+        l2_el_rpc_endpoint: &Url,
+        l2_cl_rpc_endpoint: &Url,
+        network_id: u32,
+        contracts: &AggchainProofContractsConfig,
+    ) -> Result<Self, crate::Error> {
```
Freyskeyd, line 138: Does the different RPCs shouldn't be part of the config?

Freyskeyd (approved):
Minor comment, overall LGTM

crates/aggchain-proof-contracts/src/lib.rs
atanmarko: @Freyskeyd That was my dilemma. They are already part of the config for the `aggchain-proof-builder`, didn't want to duplicate them in the configuration hence this approach.

crates/aggchain-proof-contracts/src/contracts.rs
atanmarko: That was also my dilemma, I guess you are right, It shouldn't be on the trait.

crates/aggchain-proof-contracts/src/lib.rs
atanmarko: Since they are actually not directly used by the `aggchain-proof-builder`, I'll remove them from its configuration and put them only in the `AggchainContractsClient` configuration.

iljakuklic (approved):

Issue #48, opened 2025-02-12
Implement the retrieval of the `aggregation_vkey`


atanmarko:
## Description

The `fep-proposer` is generating a `SP1ProofWithPublicInputs` that should be a `compressed` proof. This compressed proof contains the vKey used to generate the proof.

This proof needs to be used in the SP1Stdin, sent to the provin network.
It also needs to be hash with `hash_u32` to be pass to the prover

PR #83, opened 2025-03-04
feat: check aggregated fep vkey

iljakuklic:
This checks the aggregated vkey to have the expected value as specified in the config.

Related:
* #48

## PR Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added or updated tests that comprehensively prove my change is effective or that my feature works


crates/aggchain-proof-builder/src/config.rs
```diff
@@ -41,3 +47,7 @@ impl Default for AggchainProofBuilderConfig {
 fn default_aggchain_prover_timeout() -> Duration {
     Duration::from_secs(3600)
 }
+
+fn default_aggregated_proof_vkey_hash() -> B256 {
+    hex!("00441b614a713401ad1090cd7e59813e07352cb247172934f5a05dedb9e671bf").into()
```
Freyskeyd, line 52: Is it the real default value?

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -61,6 +63,27 @@ pub struct AggchainProofBuilderRequest {
...
 
+impl AggchainProofBuilderRequest {
+    /// Extract the proof and check the vkey is the expected one.
+    fn agg_span_proof_vkey(&self, expected_vkey_hash: B256) -> Result<SP1VerifyingKey, Error> {
+        let sp1_proof = &self.agg_span_proof.proof;
+        let proof = &**sp1_proof
+            .try_as_compressed_ref()
+            .ok_or_else(|| Error::UnsupportedAggProofMode(sp1_proof.into()))?;
+
+        let vk = proof.vk.clone();
```
Freyskeyd, line 74: Why do we need to clone the `vk` here?

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -61,6 +63,27 @@ pub struct AggchainProofBuilderRequest {
     pub ger_inclusion_proofs: HashMap<String, InclusionProof>,
 }
 
+impl AggchainProofBuilderRequest {
+    /// Extract the proof and check the vkey is the expected one.
+    fn agg_span_proof_vkey(&self, expected_vkey_hash: B256) -> Result<SP1VerifyingKey, Error> {
```
Freyskeyd, line 68:
```suggestion
    fn check_aggspan_proof_vkey(&self, expected_vkey_hash: B256) -> Result<(), Error> {
```

Here, we only care about knowing if the vKey of the proof matches the one in our config.

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -111,45 +202,26 @@ impl AggchainProofBuilder<AggchainContractsRpcClient<AlloyFillProvider>> {
...
-            .get_l2_output_at_block(request.end_block)
-            .await
-            .map_err(Error::L2ChainDataRetrievalError)?;
-
-        let _rollup_config_hash = contracts_client
-            .get_rollup_config_hash()
-            .await
-            .map_err(Error::L1ChainDataRetrievalError)?;
+        let _agg_span_proof_vkey =
+            request.agg_span_proof_vkey(expected_agg_span_proof_vkey_hash)?;
```
Freyskeyd, line 224: This should be done before everything else

crates/aggchain-proof-builder/src/config.rs
```diff
@@ -11,6 +12,10 @@ pub struct AggchainProofBuilderConfig {
     /// ID of the network for which the proof is generated (rollup id).
     pub network_id: u32,
 
+    /// Hash of the aggregated execution proof verification key.
+    #[serde(default = "default_aggregated_proof_vkey_hash")]
+    pub aggregated_proof_vkey_hash: B256,
```
Freyskeyd, line 17: A better place for this could be in the `proposer-service`, this service is responsible for requesting and returning a valid aggspan-proof, thus checking the vKey makes more sense in this component.

Freyskeyd (changes requested):

crates/aggchain-proof-builder/src/config.rs
atanmarko: Make sense. If the vkey is invalid, there is not point making the `aggchain-proof-builder` request to generate a proof.
Then we would need to move this expected vkey config parameter and `check_aggspan_proof_vkey` to the proposer service, and `agghcain-proof-builder` would just proceed with the agghain proof generation.

crates/proposer-service/src/error.rs
```diff
@@ -7,4 +8,10 @@ pub enum Error {
 
     #[error("Proposer client error: {0}")]
     Client(#[from] ProposerClientError),
+
+    #[error("Unsupported aggregated execution proof mode {0:?}")]
```
atanmarko, line 12:
```suggestion
    #[error("Unsupported aggregated span proof mode {0:?}")]
```

crates/proposer-service/src/error.rs
```diff
@@ -7,4 +8,10 @@ pub enum Error {
...
+    UnsupportedAggProofMode(sp1_sdk::SP1ProofMode),
+
+    #[error("Aggregated execution proof vkey mismatch (got: {got:?}, expected: {expected:?})")]
```
atanmarko, line 15:
```suggestion
    #[error("Aggregated span proof vkey mismatch (got: {got:?}, expected: {expected:?})")]
```

crates/proposer-service/src/lib.rs
```diff
@@ -19,14 +20,19 @@ mod tests;
 
 pub struct ProposerService<L1Rpc, ProposerClient> {
     pub client: Arc<ProposerClient>,
+
     pub l1_rpc: Arc<L1Rpc>,
+
+    /// Expected aggregation proof verification key.
```
atanmarko, line 26:
```suggestion
    /// Expected aggregated span proof verification key.
```

crates/proposer-service/src/lib.rs
```diff
@@ -19,14 +20,19 @@ mod tests;
...
+    aggspan_vkey_hash: B256,
```
atanmarko, line 27:
```suggestion
    agg_span_proof_vkey_hash: B256,
```

crates/proposer-service/src/config.rs
```diff
@@ -10,13 +11,22 @@ pub struct ProposerServiceConfig {
     /// JSON-RPC endpoint of the l1 node.
     #[serde(default = "prover_alloy::default_l1_node_url")]
     pub l1_rpc_endpoint: Url,
+
+    /// Hash of the aggregated execution proof verification key.
+    #[serde(default = "default_aggregated_proof_vkey_hash")]
+    pub aggregated_proof_vkey_hash: B256,
```
atanmarko, line 17:
```suggestion
    /// Hash of the aggregated span proof verification key.
    /// It verifies the proposer provided aggregated span proof (aggregated full execution proof for the block span).
    #[serde(default = "default_aggregated_proof_vkey_hash")]
    pub agg_span_proof_vkey_hash: B256,
```

crates/proposer-service/src/lib.rs
```diff
@@ -35,17 +41,39 @@ impl<L1Rpc> ProposerService<L1Rpc, proposer_client::Client<ProposerRpcClient, Ne
...
+    sp1_proof: &SP1ProofWithPublicValues,
+    expected_vkey_hash: B256,
+) -> Result<(), Error> {
+    let sp1_proof = &sp1_proof.proof;
+    let proof = &**sp1_proof
+        .try_as_compressed_ref()
+        .ok_or_else(|| Error::UnsupportedAggProofMode(sp1_proof.into()))?;
+
+    let vk = &proof.vk;
+    let vkey_hash = B256::from(vk.hash_bytes());
```
hadjiszs, line 68: Maybe using [hash_u32()](https://github.com/succinctlabs/sp1/blob/dev/crates/prover/src/types.rs#L47-L48) in all the places where we hash the aggregation vkey so that we could use the same functions to also check that it is properly equal to the `AGGREGATION_VKEY_HASH` hardcoded in `aggchain-proof-core` [here](https://github.com/agglayer/provers/blob/ff47b25d255bc0738154c316bf6383a71c06d7c1/crates/aggchain-proof-core/src/full_execution_proof.rs#L11-L13)

iljakuklic: Ok. will switch to `hash_u32`.

Regarding the vkey value, I introduced a config value (with a default) for this. Does it make sense for it to be a config value or should it be just a constant hardcoded in code? If it's in config, it would have to be passed into the proof. It probably does not make sense to have both. Thoughts? @atanmarko @Freyskeyd

crates/proposer-service/src/vkey_hash.rs
```diff
@@ -0,0 +1,106 @@
+use alloy_primitives::B256;
+use serde::{Deserialize, Serialize};
+use sp1_sdk::HashableKey;
+
+pub type HashU32 = [u32; 8];
+
+/// SP1 verifying key hash.
+#[derive(Clone, Copy, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+#[serde(from = "B256", into = "B256")]
+pub struct VKeyHash(HashU32);
```
iljakuklic, line 10: Added this type so we can use `VKeyHash`es consistently. It is now in a submodule here but really it should go into some sort of common crate so other parts of code base have access to it. There is nothing specific to the proposer service in this module.

Ekleog-Polygon (commented):
Overall LGTM! Just have a few comments on the VKeyHash conversion functions, but I'll leave the final r+ to someone with better knowledge of what we're trying to do for now :)

crates/proposer-service/src/lib.rs
```diff
@@ -35,17 +42,38 @@ impl<L1Rpc> ProposerService<L1Rpc, proposer_client::Client<ProposerRpcClient, Ne
...
             )?),
+            agg_span_proof_vkey_hash,
         })
     }
 }
 
+fn check_aggspan_proof(
+    sp1_proof: &SP1ProofWithPublicValues,
+    expected_vkey_hash: VKeyHash,
+) -> Result<(), Error> {
```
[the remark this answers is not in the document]
iljakuklic: > Or maybe this function should be called `check_aggspan_proof_vk`, to make it clear that we're not actually checking the proof?

Yea, the name is misleading, will update

crates/proposer-service/src/vkey_hash.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        let mut w = 0_usize;
+        while w < 8 {
+            let b0 = bytes[4 * w];
+            let b1 = bytes[4 * w + 1];
+            let b2 = bytes[4 * w + 2];
+            let b3 = bytes[4 * w + 3];
+            hash_u32[w] = u32::from_be_bytes([b0, b1, b2, b3]);
+            w += 1;
+        }
```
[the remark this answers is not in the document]
iljakuklic: I appreciate the implementation is kind of ugly but I wanted to stick to stuff that would allow me to make this a `const fn`, so we can make a compile time constants of this type like so:
```rust
const SOME_VKEY_HASH: VKeyHash: VKeyHash::from_bytes(b256!("0xSOMEMAGICNUMBER"));
```

But maybe the other direction does not need to be a `const fn`.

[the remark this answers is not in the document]
Ekleog-Polygon: Oh I had missed this needed to be a const fn. Sorry for the noise then! :)

crates/proposer-service/src/lib.rs
```diff
@@ -35,17 +42,38 @@ impl<L1Rpc> ProposerService<L1Rpc, proposer_client::Client<ProposerRpcClient, Ne
...
+        .try_as_compressed_ref()
+        .ok_or_else(|| Error::UnsupportedAggProofMode(sp1_proof.into()))?;
+
+    let vkey_hash = VKeyHash::from_vkey(&proof.vk);
+    (vkey_hash == expected_vkey_hash)
+        .then_some(())
+        .ok_or(Error::AggProofVKeyMismatch {
+            got: vkey_hash,
+            expected: expected_vkey_hash,
+        })
```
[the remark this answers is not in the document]
iljakuklic: Changed. Would be nice to have an equivalent of `anyhow::ensure!` for typed errors, this is a very common pattern.

crates/proposer-service/src/vkey_hash.rs
atanmarko: Maybe you could put it immediately to the `aggkit-prover-types`? There is some `pub type Hash` already there.

crates/proposer-service/src/vkey_hash.rs
```diff
@@ -0,0 +1,107 @@
+use alloy_primitives::B256;
+use serde::{Deserialize, Serialize};
+use sp1_sdk::HashableKey;
+
+pub type HashU32 = [u32; 8];
```
atanmarko, line 5: Why is this hash array of u32 and not array of u8? I see that Sp1 `HashableKey` has `bytes32_raw()`, do we need to use `hash_u32()`? It would make conversions  more simple.

iljakuklic: > Why is this hash array of u32 and not array of u8? I see that Sp1 `HashableKey` has `bytes32_raw()`, do we need to use `hash_u32()`? It would make conversions more simple.

I'm not quite sure these are guaranteed to be the same...

atanmarko: What about 
```
    /// Hash the key into a digest of bytes elements.
    fn hash_bytes(&self) -> [u8; DIGEST_SIZE * 4] {
        words_to_bytes_be(&self.hash_u32())
    }
```

iljakuklic: Also the idea is that eventually, this might also be used in the context of using the vkey to the proof where the `[u32; 8]` format is expected. It is also currently used in @hadjiszs's PR where the aggregated vkey hardcoded constant is defined. In general, I'd say it's better to keep closer to the "processed" representation and parse raw data at the boundaries.

iljakuklic:
Putting into draft while waiting how #94 will pan out.

iljakuklic:
Waiting on #84 to get access to the vkey hash constant.

crates/aggchain-proof-builder/src/config.rs
iljakuklic: The vkey is now a constant and it just refers to the value from `aggchain_proof_core`.

iljakuklic:
> Waiting on #84 to get access to the vkey hash constant.

Done.

atanmarko (dismissed):

atanmarko (dismissed):

Freyskeyd (dismissed):

crates/proposer-service/src/lib.rs
```diff
@@ -17,16 +18,24 @@ pub mod error;
...
+const AGGREGATION_VKEY_HASH: VKeyHash =
+    VKeyHash::from_hash_u32(aggchain_proof_core::AGGREGATION_VKEY_HASH);
+
 pub struct ProposerService<L1Rpc, ProposerClient> {
     pub client: Arc<ProposerClient>,
+
     pub l1_rpc: Arc<L1Rpc>,
+
+    /// Expected aggregated span proof verification key.
+    aggregation_vkey_hash: VKeyHash,
```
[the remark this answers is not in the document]
iljakuklic: It is initialized with the constant. It is (for now) used in tests too to check the expected behaviour. It is difficult to do it otherwise since obtaining a vkey that hashes to `[0u32; 8]` is a bit tricky.

Freyskeyd (approved):

PR #94, opened 2025-03-11
refactor: introduce the `VKeyHash` type

iljakuklic:
* extracted from #83

## PR Checklist:

### DIFF
--- a/crates/aggkit-prover-types/src/vkey_hash.rs
+++ b/crates/aggkit-prover-types/src/vkey_hash.rs
@@ -1,0 +1,107 @@
+use alloy_primitives::B256;
+use serde::{Deserialize, Serialize};
+
+pub type HashU32 = [u32; 8];
+
+/// SP1 verifying key hash.
+#[derive(Clone, Copy, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+#[serde(from = "B256", into = "B256")]
+pub struct VKeyHash(HashU32);
+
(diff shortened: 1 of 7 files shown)


crates/aggkit-prover-types/src/vkey_hash.rs
iljakuklic, line 1: Seems to me it would be even more useful to place this in some crate that could be shared with `aggchain-proof-core` because it manipulates vkey hashes and could make use of this.

Or maybe putting it into `aggchain-proof-core` directly would work for now? Seems a bit inappropriate since the type is not specific to aggchain proof.

atanmarko: As `Digest` is in the `aggchain-proof-core`, you can put vkey there as well.
We can reexport it in the `agghcain-proof-types` like [here](https://github.com/agglayer/provers/pull/87/files).

crates/aggkit-prover-types/src/vkey_hash.rs
```diff
@@ -0,0 +1,107 @@
+use alloy_primitives::B256;
+use serde::{Deserialize, Serialize};
+
+pub type HashU32 = [u32; 8];
```
atanmarko, line 4: Per discussion, this should be moved to the `aggchain-proof-core`.

atanmarko (commented):
LGTM

atanmarko (approved):

PR #98, opened 2025-03-12
build: make docker compile and update aggkit binary

Freyskeyd:
# Description

This PR is refactoring the dockerfile and how the aggkit binary is starting.

## PR Checklist:


crates/aggkit-prover/src/cli/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+    },
+
+    Config,
+
+    ValidateConfig {
+        /// The path to the aggkit-prover configuration file.
+        #[arg(
+            value_hint = ValueHint::DirPath,
+        )]
+        config_path: PathBuf,
```
iljakuklic, line 31: Is this a directory or a file? The hint does not agree with the comment.

Also a nitL does the `#[arg(..)]` attribute need to be over three lines?

crates/agglayer-prover/src/lib.rs
```diff
@@ -51,9 +51,10 @@ pub fn main(cfg: PathBuf, version: &str, program: &'static [u8]) -> anyhow::Resu
 
     Ok(())
 }
+
 pub fn get_vkey(program: &'static [u8]) -> String {
     let vkey = prover_executor::Executor::get_vkey(program);
-    vkey.bytes32().to_string()
+    vkey.bytes32()
 }
```
iljakuklic, line 58: Is this function even used? I don't think it is in our codebase but I see it's public so external dependents may use it. Still I am not sure it provides much value.

crates/aggkit-prover/src/main.rs
```diff
@@ -1,46 +1,47 @@
...
+            }
+        }
+        aggkit_prover::cli::Commands::Vkey => {
+            let vkey =
+                prover_executor::Executor::get_vkey(aggchain_proof_service::AGGCHAIN_PROOF_ELF);
+
+            println!(
+                "aggchain_proof_vkey: 0x{}",
+                hex::encode(words_to_bytes_le(&vkey.hash_u32()))
+            );
```
iljakuklic, line 42: Nit: I think the `println!`s are a bit nicer with variables:
```suggestion
            let vkey_hex = hex::encode(words_to_bytes_le(&vkey.hash_u32()));
            println!("aggchain_proof_vkey: 0x{vkey_hex}");
```

There are two more above.

crates/aggkit-prover/src/cli/mod.rs
Freyskeyd: It is the path to the configuration file

crates/aggkit-prover/src/cli/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+        #[arg(long, short, value_hint = ValueHint::FilePath, default_value = "aggkit-prover.toml", env = "CONFIG_PATH")]
+        config_path: PathBuf,
+    },
+
+    Config,
+
+    ValidateConfig {
+        /// The path to the aggkit-prover configuration file.
+        #[arg(value_hint = ValueHint::DirPath)]
+        config_path: PathBuf,
```
iljakuklic, line 29: My point in one of the previous comments was that there is a mismatch. The comment mentions "configuration *file*" while the value hint is a `DirPath` rather than `FilePath`.

Freyskeyd: Oh I see !
Good catch, I didn't understand the previous comment that way, thanks for the clarification !

iljakuklic (approved):

PR #68, opened 2025-02-25
feat: not inclusive start block

invocamanman:
# Description

We use the `start_block` as the first block that's proved. That would mean that the block proven follows the ` [start_block, finalBlock]` convention. 

Usually in rollups is used the `(start_block, finalBlock]` convention, so, start block not inclusive.
This is useful usually to check that the last `finalBlock` matches the `start_block` of the next proving block.

As an example of our current and new approach given a scenario with 2 range proofs. The first proof proves the block 1 and 2 and the second one proves the block 3:

## Current Approach

| Proof    | `start_block` | `end_block` |
|----------|---------------|-------------|
| Proof 1  | 1             | 2           |
| Proof 2  | 3             | 3           |

---

## New Convention

| Proof    | `start_block` | `end_block` |
|----------|---------------|-------------|
| Proof 1  | 0             | 2           |
| Proof 2  | 2             | 3           |

## Additions and Changes

I fix a condition and add comments regarding this change
Mainly i think this will be relevant in the upcoming PRs. 

For example now instead of doing this kinda[ minus one scenarios](https://github.com/agglayer/provers/blob/086e7614670fb49330af3d66397f3add15893498/crates/aggchain-proof-builder/src/lib.rs#L119) we will be able to use `start_block` right away

It's similar to the common convention with the rollups and personally i find it more intuitive ( maybe bc i work on a rollup for a while :P) 

## PR Checklist:

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added or updated tests that comprehensively prove my change is effective or that my feature works

PR #122, opened 2025-03-26
feat: adjust start/end block properly

Ekleog-Polygon:
Closes https://github.com/agglayer/agglayer/issues/731
Closes #68 

## PR Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation


crates/proposer-client/src/rpc.rs
```diff
@@ -28,14 +28,17 @@ pub trait AggregationProofProposer {
...
 #[serde(rename_all = "camelCase")]
 pub struct AggregationProofProposerRequest {
-    /// Starting block number to request proof from.
+    /// Last block that has already been proven before this request.
     #[serde(rename = "startBlock")]
-    pub start_block: u64,
+    pub last_proven_block: u64,
+
     /// Maximum block number for the aggregation proof.
     #[serde(rename = "maxBlock")]
```
hadjiszs, line 36: the serde rename shouldnt be adjusted too?

hadjiszs (approved):
lgtm! happy to get a double check from @invocamanman 

my only doubt was on the `EVMStateSketch` tied to the starting block but I think that's correct: we want to get the hash chains from the `last_proven_block`, and then the events should be from the block `last_proven_block + 1` as defined by the range `new_blocks_range` on which we filter the claims/inserted gers etc

crates/proposer-client/src/rpc.rs
Ekleog-Polygon: Great point, thank you! I just did a second pass, as I had completely missed these annotations

hadjiszs (approved):

iljakuklic (commented):
Looks good to me. I am not 100% sure where all the +/-1 adjustments ought to be though.

scripts/make/Makefile.proto.toml
```diff
@@ -0,0 +1,57 @@
...
+[tasks.generate-aggkit-proto]
+command = "buf"
+args = ["generate", "--template", "buf.aggkit-prover.gen.yaml"]
+dependencies = [
+    "install-buf",
+    "install-protoc-gen-prost",
+    "install-protoc-gen-prost-serde",
+    "install-protoc-gen-prost-crate",
+    "install-protoc-gen-tonic",
+]
```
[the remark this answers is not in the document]
Ekleog-Polygon: Will do in a follow-up PR 😄

PR #123, opened 2025-03-26
ci: enable docker build on `main` push and on manual trigger

iljakuklic:
## PR Checklist:


.github/workflows/release.yml
```diff
@@ -56,8 +61,8 @@ jobs:
       - name: Image name builder
         id: image_builder
         run: |
-          IMAGE=$(jq -ecr '.tags | map(select(match("${{ env.REGEX_IMAGE }}", "i"))) | first| sub(":.*$";"")' <<< "$DOCKER_METADATA_OUTPUT_JSON")
-          echo "IMAGE=$IMAGE" >> $GITHUB_OUTPUT
+          IMAGE=$(jq -ecr '.tags | map(select(match("${{ env.REGEX_IMAGE }}"; "i"))) | first| sub(":.*$";"")' <<< "$DOCKER_METADATA_OUTPUT_JSON")
+          echo "IMAGE=$IMAGE" | tee $GITHUB_OUTPUT
```
iljakuklic, line 65: This was a fun one. The comma creates a list, which the `match` function of `jq` treats as a list of patterns to match. So we were always matching the "i" character in the "ghcr.io" part of the image tag. Took me a while to figure out why the workflow works without updating `REGEX_IMAGE` even though it should not.

Ekleog-Polygon (commented):
Can we just copy-paste the whole .github folder from agglayer to provers? I'm pretty sure the docker builds are not the only thing missing, and we spent quite a while polishing agglayer CI.

(For later we should likely make a separate repo and just workflow_call into it so we're not duplicating CI on both ends, but the copy-paste should be good enough for now hopefully)

Ekleog-Polygon (approved):
Actually I'll approve because the diff looks safe to me, but I still think we should bring in everything from agglayer rather than just let the two CI require manual handholding once each

iljakuklic:
> Can we just copy-paste the whole .github folder from agglayer to provers? I'm pretty sure the docker builds are not the only thing missing, and we spent quite a while polishing agglayer CI.
> 
> (For later we should likely make a separate repo and just workflow_call into it so we're not duplicating CI on both ends, but the copy-paste should be good enough for now hopefully)

I don't think it's as trivial as just copy and paste. For exmaple, the agglayer workflow inserts the e2e tests between image build and image push. It probably would not work out of the box.

I'd say it's better to just do it properly in the separate repo without the copy-paste middle step but that requires some time and effort.

hadjiszs (approved):

PR #130, opened 2025-03-28
fix: service ready

atanmarko:
Fix the call of the prover executor.


crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -306,6 +307,9 @@ where
                 Self::retrieve_chain_data(contracts_client, req, network_id).await?;
 
             let result = prover
+                .ready()
+                .await
+                .map_err(Error::ProverServiceReadyError)?
```
iljakuklic, line 312: Not quite related to this PR but I just noticed the `Err` result form the call is currently handled with a `todo!()`.

iljakuklic (approved):
+1

iljakuklic (approved):

PR #131, opened 2025-03-31
feat: make aggkit-prover startup faster by 2x

### LINKED ISSUE #126: Aggkit prover takes too much time to start

Ekleog-Polygon:
Startup goes from ~14s to ~7s on my machine.

The time is heavily dominated by building the SP1 provers: each prover build takes ~3s. This PR removes two prover builds, and only leaves two: one for the aggregation elf, and one for the aggchain proof elf.

We might be able to shave another 3s (so another 2x factor) by only having a single prover, but it would be harder to implement, and is thus left as future work for now.

Fixes #126

## PR Checklist:

### DIFF
--- a/crates/agglayer-prover/src/lib.rs
+++ b/crates/agglayer-prover/src/lib.rs
@@ -1,7 +1,6 @@
 use std::{path::PathBuf, sync::Arc};
 
 use prover_engine::ProverEngine;
-use sp1_sdk::HashableKey;
 
 #[cfg(feature = "testutils")]
 pub mod fake;
@@ -50,11 +49,6 @@ pub fn main(cfg: PathBuf, version: &str, program: &'static [u8]) -> anyhow::Resu
         .start();
 
     Ok(())
-}
-
-pub fn get_vkey(program: &'static [u8]) -> String {
-    let vkey = prover_executor::Executor::get_vkey(program);
-    vkey.bytes32()
 }
 
 #[cfg(feature = "testutils")]
--- a/crates/prover-executor/src/lib.rs
+++ b/crates/prover-executor/src/lib.rs
@@ -27,16 +27,14 @@ mod error;
 
 #[derive(Clone)]
 pub struct Executor {
+    vkey: SP1VerifyingKey,
     primary: BoxCloneService<Request, Response, Error>,
     fallback: Option<BoxCloneService<Request, Response, Error>>,
 }
 
 impl Executor {
-    pub fn get_vkey(elf: &[u8]) -> SP1VerifyingKey {
-        let prover = CpuProver::new();
-        let (_proving_key, verification_key) = prover.setup(elf);
-
-        verification_key
+    pub fn get_vkey(&self) -> SP1VerifyingKey {
+        self.vkey.clone()
     }
 
     pub fn build_network_service<S>(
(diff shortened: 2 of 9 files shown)


atanmarko (approved):

crates/agglayer-prover/src/lib.rs
```diff
@@ -52,11 +51,6 @@ pub fn main(cfg: PathBuf, version: &str, program: &'static [u8]) -> anyhow::Resu
     Ok(())
 }
 
-pub fn get_vkey(program: &'static [u8]) -> String {
-    let vkey = prover_executor::Executor::get_vkey(program);
-    vkey.bytes32()
-}
-
```
Freyskeyd, line 59: Why this public function is removed? [It is used here](https://github.com/agglayer/agglayer/blob/main/crates/agglayer/src/main.rs#L42)

Freyskeyd: I would prefer to keep the logic encapsulated into the `prover` crate, the CLI shouldn't have to have `sp1-sdk` or even `prover-executor` as direct dep.
I'm fine refactoring this to a standalone function, tho. Something like:

```rust

// prover_executor crate
pub fn get_program_vkey(program: &'static [u8]) -> String { ... }

// agglayer_prover

pub fn get_vkey(program: &'static [u8]) -> String {
    let vkey = prover_executor::get_program_vkey(program);
    vkey.bytes32()
}
```

crates/prover-executor/src/lib.rs
```diff
@@ -27,16 +27,14 @@ mod error;
 
 #[derive(Clone)]
 pub struct Executor {
+    vkey: SP1VerifyingKey,
```
Freyskeyd, line 30:
```suggestion
    vkey: Arc<SP1VerifyingKey>,
```

Do you think we could `arc` this?

crates/agglayer-prover/src/lib.rs
```diff
@@ -52,8 +51,8 @@ pub fn main(cfg: PathBuf, version: &str, program: &'static [u8]) -> anyhow::Resu
     Ok(())
 }
 
-pub fn get_vkey(program: &'static [u8]) -> String {
-    let vkey = prover_executor::Executor::get_vkey(program);
+pub fn compute_program_vkey(program: &'static [u8]) -> String {
```
Freyskeyd, line 54: We'll need to update the agglayer node for this.
We should also start tagging the agglayer-node's prover deps I think

Freyskeyd (approved):

crates/agglayer-prover/src/lib.rs
Ekleog-Polygon: 👍 I'll handle that right now 😄

And I'm not entirely sure about tagging the deps — for so long as we have aggkit-prover mixed in here, tagging this repo doesn't really make sense, and tagging each dep independently will be a mess. So maybe just relying on the `Cargo.lock` makes most sense for now, and as soon as we get this repo actually only contain shareable code we can handle releases properly?

BTW I'm also thinking of setting up [release-plz](https://github.com/release-plz/release-plz) on all our repos, will probably look into it at some point if I ever have time to look into our CI status.

PR #132, opened 2025-04-01
feat: add execute program test

### LINKED ISSUE #133: Implement test to run prover executor with aggchain prover inputs

atanmarko:
# Description

Added test with execution of the aggchain proof program.

Fixes #133 

## PR Checklist:


crates/aggchain-proof-builder/src/tests/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+                stdin: aggchain_prover_inputs.stdin,
+                proof_type: prover_executor::ProofType::Stark,
+            })
+            .await
+            .map_err(|error| Error::ProverFailedToExecute(anyhow::Error::from_boxed(error)))?;
+
+        println!(
+            "Prover executor successfully returned response: {:?} ",
+            proof
+        );
```
iljakuklic, line 75: Could this be a snapshot instead of a printout?

iljakuklic (approved):

crates/aggchain-proof-builder/src/tests/mod.rs
atanmarko: @iljakuklic What do you mean?

iljakuklic: > @iljakuklic What do you mean?

something like:
```suggestion
        insta::debug_snapshot!(proof);
```

So it writes the result to a file that is seen in the repo. I don't know if it's a good idea to keep in the file. For example, it requires the output to be deterministic so not good if it contains hashtables and such. Just a random suggestion to consider.

atanmarko: My idea of printout was just to note that the test  passed, not really to look into result (which is just a byte array).  We can remove printout as well :)

PR #135, opened 2025-04-01
fix: align vk and public value mismatch

hadjiszs:
# Description

- [x] fix aggregation vkey mismatch by retrieving it from the committed ELF
- [x] redo public values encoding also for easier comparison with public values coming from the proof

## PR Checklist:


iljakuklic (approved):
Looks good.

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -253,13 +265,33 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
             },
         };
 
-        let aggregation_proof = request.aggregation_proof;
-        let aggregation_vkey = aggregation_proof.vk.clone();
+        // Retrieve the entire aggregation vkey from the ELF
+        let aggregation_vkey = {
+            let prover = ProverClient::builder().cpu().build();
+            let (_, agg_vk_from_elf) = prover.setup(AGGREGATION_PROOF_ELF);
+            agg_vk_from_elf
```
Ekleog-Polygon, line 272: This will take 3 seconds on each run. Why not reuse the one we already have computed? See https://github.com/agglayer/provers/pull/131 for more details

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -253,13 +265,33 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
+                    expected,
+                });
+            }
+        }
+
         let sp1_stdin = {
             let mut stdin = SP1Stdin::new();
             stdin.write(&prover_witness);
-            stdin.write_proof(*aggregation_proof, aggregation_vkey);
+            stdin.write_proof(*aggregation_proof, aggregation_vkey.vk);
```
Ekleog-Polygon, line 294: AFAIU this is all computed once. Can we at least wrap it in a `OnceCell` so that we're computing it a single time per program run, and not once per RPC call? It'd be even better if we can reuse the vkey already computed in #131 that said.

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -253,13 +265,33 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
+
+        let aggregation_proof = request.aggregation_proof.clone();
         let witness = prover_witness.clone();
+
+        // Check mismatch on aggregation vkey
+        {
+            let retrieved = VKeyHash::from_vkey(&aggregation_vkey);
+            let expected = VKeyHash::from_hash_u32(AGGREGATION_VKEY_HASH);
+
+            if retrieved != expected {
```
Ekleog-Polygon, line 283: AFAIU this sanity-check only uses constant values. Can we just call it once per program startup, and not once per `retrieve_chain_data` call?

Ekleog-Polygon (changes requested):
There are quite a few things that look to me like they don't belong in the function where they are, unless I missed something?

crates/aggchain-proof-builder/src/lib.rs
atanmarko: Yeah, this is totally unoptimal, I'm making the new pr

atanmarko: Fixed with https://github.com/agglayer/provers/pull/139

crates/aggchain-proof-builder/src/lib.rs
atanmarko: Ok, will implement in the next PR

crates/aggchain-proof-builder/src/lib.rs
atanmarko: Fixed in https://github.com/agglayer/provers/pull/139

PR #139, opened 2025-04-02
fix: optimize vkey generation

atanmarko:
# Description

Fix vkey generation

### DIFF
--- a/crates/aggchain-proof-builder/src/lib.rs
+++ b/crates/aggchain-proof-builder/src/lib.rs
@@ -103,7 +103,7 @@ pub struct AggchainProofBuilder<ContractsClient> {
     prover: ProverService,
 
     /// Verification key for the aggchain proof.
-    aggchain_proof_vkey: Arc<SP1VerifyingKey>,
+    aggregation_vkey: Arc<SP1VerifyingKey>,
 }
 
 #[derive(Debug, Clone, thiserror::Error)]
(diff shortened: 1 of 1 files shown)


crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -122,17 +122,37 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
         );
-        let aggchain_proof_vkey = executor.get_vkey().clone();
 
         let executor = tower::ServiceBuilder::new().service(executor).boxed();
 
         let prover = Buffer::new(executor, MAX_CONCURRENT_REQUESTS);
 
+        // Retrieve the entire aggregation vkey from the ELF
+        let aggregation_vkey = {
+            // TODO, use executor.get_vkey().clone() instead, just test before
```
Ekleog-Polygon, line 132: Unfortunately I don't think you could use that here, because it's not the same program.

However, do you know if a prover can be setup for multiple ELFs? If yes, you could shave an additional 3s by adding an `executor.compute_program_vkey()` that'd reuse the already-setup prover and compute the vkey using it

Ekleog-Polygon, line 132: Unfortunately I don't think you could use that here, because it's not the same program.

However, do you know if a prover can be setup for multiple ELFs? If yes, you could shave an additional 3s at startup by adding an `executor.compute_program_vkey()` that'd reuse the already-setup prover and compute the vkey using it.

This being said, running this only once each boot should be fine. Thank you for the PR!

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -122,17 +122,37 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
+            let (_, agg_vk_from_elf) = prover.setup(AGGREGATION_PROOF_ELF);
+            agg_vk_from_elf
+        };
+
+        // Check mismatch on aggregation vkey
+        {
+            let retrieved = VKeyHash::from_vkey(&aggregation_vkey);
+            let expected = VKeyHash::from_hash_u32(AGGREGATION_VKEY_HASH);
+
+            if retrieved != expected {
```
Ekleog-Polygon, line 143: For full disclosure, I'm even thinking we could maybe serialize the aggregation proof vkey in the build.rs and just deserialize it at runtime, but maybe that'd be too much

Ekleog-Polygon (approved):

crates/aggchain-proof-builder/src/lib.rs
iljakuklic: > For full disclosure, I'm even thinking we could maybe serialize the aggregation proof vkey in the build.rs and just deserialize it at runtime, but maybe that'd be too much

I'm planning to do something like this as a part of #138

PR #174, opened 2025-04-16
feat: aggchain params encoding

hadjiszs:
# Description

Rearrange encoding based on the one in the contracts

## PR Checklist:


crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -286,6 +289,20 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
+
+            info!("Aggchain-params unrolled values: {aggchain_params_values:?}");
+            info!(
+                "Aggchain-params abi encoded packed: {}",
+                hex::encode(fep_inputs.encoded_aggchain_params())
+            );
+            info!(
+                "Aggchain-params keccak-hashed: {}",
+                fep_inputs.aggchain_params()
+            );
```
atanmarko, line 303: This will make a lot of noise in the logs. Should it be `trace`?

crates/aggchain-proof-program/Cargo.lock
iljakuklic, line 1: Curious why did this change without any update in `Cargo.toml`? Any of the transitive dependencies needed updating?

crates/aggchain-proof-core/src/full_execution_proof.rs
```diff
@@ -79,25 +104,20 @@ impl FepInputs {
...
-
-        keccak256_combine([
-            self.compute_l2_pre_root().as_slice(),
-            self.compute_claim_root().as_slice(),
-            &self.claim_block_num.to_be_bytes(),
-            self.rollup_config_hash.as_slice(),
-            &optimistic_mode.to_be_bytes(),
-            self.trusted_sequencer.as_slice(),
-        ])
+        keccak256(self.encoded_aggchain_params().as_slice())
```
iljakuklic, line 120: Not strictly related to the PR but I think `aggchain_params_commitment` would be a more accurate name.

crates/aggchain-proof-builder/src/lib.rs
hadjiszs: very needed for now I think so `info` sounds reasonable to me to avoid having to say to everyone to enable the right log level

iljakuklic (approved):

crates/aggchain-proof-core/src/full_execution_proof.rs
hadjiszs: ah `aggchain_params` as-is really refers to the hash itself same as in the specs and in the contracts

PR #175, opened 2025-04-17
feat: add context useful for debugging to the aggchain proof

### LINKED ISSUE #146: Adding the `public values` generated by the `aggchain-proof` to the `context` returned by the `aggkit-prover`
# Description

In order to ease the debugging, we should provide a serialize version of the `public values` as part of the `aggchain-proof`'s context.

The `aggchain-proof` contains a `map<string, bytes>` that can store this value.

The agglayer will then deserialize it and add it as tracing field if needed

Ekleog-Polygon:
Fixes #146

## PR Checklist:


iljakuklic (approved):
+1. This, however, seems like a fair amount of extra data being transmitted. We may want to have a way of turning off the context stuff eventually.

Ekleog-Polygon:
Agreed, but my guess is this should also still be negligible compared to the full proof, so there's probably no hurry 😄

PR #179, opened 2025-04-23
build: zkvm build fixes & improvements

iljakuklic:
# Description

* Introduce a binary to re-generate `op-succinct` vkeys manually. Generating them automatically in `build.rs` causes trouble when compiling to zkvm due to missing system dependencies in the docker image.
* Ditch the custom SP1 toolchain docker image in favour of `cargo prove build --docker`
* Adjust `cargo make ap-elf` to regenerate vkeys and use `cargo prove`
* Small improvements to the `ElfInfo` generator.
* Rebuild the AP elf.

Related:

* Docker AP build broken since #138
* AP elf outdated since #113

## PR Checklist:

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added or updated tests that comprehensively prove my change is effective or that my feature works

PR #183, opened 2025-04-23
feat: mv vkey hashes to aggchain params as pv

hadjiszs:
# Description

Move the range vkey commitment and the aggregation vkey hash as public inputs through the aggchain params.

Relates to https://github.com/agglayer/protocol-research/issues/140

## PR Checklist:


crates/aggchain-proof-core/src/full_execution_proof.rs
```diff
@@ -166,7 +177,7 @@ impl FepInputs {
             #[cfg(target_os = "zkvm")]
             {
                 sp1_zkvm::lib::verify::verify_sp1_proof(
-                    &AGGREGATION_VKEY_HASH,
+                    &self.aggregation_vkey_hash,
```
iljakuklic, line 180: Since the aggchain verification key no longer commits to the aggregation and range vkeys, what enforces the keys are the expected ones? Id the L1 contract now supposed to check the keys specified as public values against some sort of white list?

crates/aggchain-proof-core/Cargo.toml
```diff
@@ -7,7 +7,6 @@ edition = "2021"
 agglayer-primitives.workspace = true
 agglayer-tries.workspace = true
 unified-bridge.workspace = true
-proposer-vkeys-raw.workspace = true
```
iljakuklic, line 10: The fact that the proof no longer directly depends on these vkeys has fairly profound implication on the build dependency graph. Is this a long term change?

If so, it would allow us to simplify the build significantly:
* Since the build of `proposer-vkeys-raw` involves running `SP1Prover` to calculate the vkeys, it's build dependencies are pretty heavy and are not available in the sp1 toolchain docker image.
  * It means that the switch from `build.rs` to an external command does not need to happen as proposed in #179
  * We will be able to build the proof without the extra vkey generation step.
* The separation between `proposer-vkeys-raw` and `proposer-elfs` was introduced specifically to avoid having the proof depend on unsupported stuff.
  * The two crates can be merged if the proof no longer depends on vkey constants.

Related:
* #179 
* #138

iljakuklic (commented):
Not a full review yet, just some observations. Would like to hear back regarding these comments.

crates/aggchain-proof-core/src/full_execution_proof.rs
hadjiszs: yeah exactly, they are moved as public inputs through the `aggchain_params` which is then propagated as public input of the pessimistic proof through the `aggchain_hash` which is itself re-constructed on the L1.

In other words, if the vkeys used here arent the ones registered on L1, then the `aggchain_hash` would be incorrect because of the `aggchain_params` which wont contain the right vkeys hash

Ekleog-Polygon (approved):
LGTM, though I totally agree with @iljakuklic that we will probably want to do some follow-up changes if this is here to stay

crates/aggchain-proof-core/Cargo.toml
hadjiszs: >Is this a long term change?

yeah it is thought as long term change I believe, maybe @krlosMata can confirm

iljakuklic (approved):

crates/aggchain-proof-core/Cargo.toml
krlosMata: yes, it a long term change.
During integration, we found that `op-proposer` retrieves [those parameters from the succint SC](https://github.com/succinctlabs/op-succinct/blob/main/contracts/src/validity/OPSuccinctL2OutputOracle.sol#L66-L71) and so we add them into the SC with no boundaries against the circuit ones. [This caused an inconsistency](https://github.com/agglayer/protocol-research/issues/140).

Therefore, we adapt to how `op-proposer` works and remove those hardcoded parameters from program and them via the `aggchain_params`.

Freyskeyd (approved):

crates/aggchain-proof-core/src/full_execution_proof.rs
hadjiszs: I think they won't need to be hardcoded anymore and could be dynamically fetched from L1 contracts, so basically your option 3.

For debugging I think simply logging them to know which ones were considered would be good enough

I'm spinning off this as dedicated issue to track that one 👍

hadjiszs: cross posting the issue: #189

PR #186, opened 2025-04-24
feat: integrate rust proposer

### LINKED ISSUE #169: Add mock proof retrieval support for Rust proposer
Rust proposer returns mock proof immediately with the [request_id](https://github.com/agglayer/op-succinct/blob/vcastellm/rust-agglayer/proto/proofs.proto#L24) field. It should be de-serialized, and the step where proposer-service waits for the mock proof should be removed - or even better should deserialize and return mock proof from the `request_id` byte array, to keep same execution path as for real proof where it waits for the cluster returning proof from the proof id).

atanmarko:
# Description

Add grpc and mock proof support for rust proposer

Fixes #169 

## Breaking changes

Uncompatible with go proposer


crates/proposer-client/Cargo.toml
```diff
@@ -26,16 +26,28 @@ tonic.workspace = true
 tracing.workspace = true
 url.workspace = true
 
+# Ecosystem dependencies
+op-succinct-grpc = { git = "https://github.com/agglayer/op-succinct.git", tag = "v2.1.3-agglayer", package = "grpc" }
```
iljakuklic, line 30: I am a bit conflicted about having the dependency here vs. at the workspace level:

In favour of keeping it here:
* This should be the only crate which uses `op-succinct-grpc`. All other crates should issue proof requests through this crate. So the dependency should not be globally accessible.

In favour of putting it at the workspace level:
* All dependencies on `op-succinct` would be in one place, making it easier to keep the versions in sync.

I guess we will not get automated bumps from dependabot since we depend on a specific tag.

Thoughts?

crates/aggkit-prover/src/rpc.rs
```diff
@@ -180,24 +180,9 @@ impl AggchainProofGrpcService for GrpcService {
...
-                        ),
-                    ) if e.code() == -2000 => Err(Status::unavailable(
-                        "Proposer service has not built any proof yet",
-                    )),
-                    e => Err(Status::internal(e.to_string())),
-                }
-            }
+            // TODO: Return a different error when the proof is not yet ready.
+            // The gRPC API currently does not expose the status.
+            Err(e) => Err(Status::internal(e.to_string())),
```
iljakuklic, line 185: Wondering if we can recover the custom error here. Maybe we could even propagate the status, since we now have gRPC on both ends.

If you don't want to do it now, file an issue and I'll have a look as a follow up. @atanmarko

crates/proposer-client/src/rpc/proofs_service_types.rs
```diff
@@ -0,0 +1,67 @@
...
+            last_proven_block,
+            end_block,
+        })
+    }
+}
+
+impl From<MockProofProposerRequest> for grpc::GetMockProofRequest {
+    fn from(request: MockProofProposerRequest) -> Self {
+        grpc::GetMockProofRequest {
+            proof_id: request.proof_id as i64,
```
iljakuklic, line 52: Here is the conversion I mentioned in one of the other comments. Note that it panics in debug mode and may underflow in release.

crates/proposer-client/src/mock_grpc_prover.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        Ok(proof)
+    }
+
+    fn verify_aggregated_proof(
+        &self,
+        proof: &SP1ProofWithPublicValues,
+        vkey: &SP1VerifyingKey,
+    ) -> Result<(), SP1VerificationError> {
+        self.sp1_prover.verify(proof, vkey)
```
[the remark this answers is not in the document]
atanmarko: It is a no-op I think. But lets be consistent and keep the structure of `AggregationProver`.

crates/aggkit-prover/src/rpc.rs
atanmarko: Here it is https://github.com/agglayer/provers/issues/188

iljakuklic (dismissed):

iljakuklic (dismissed):

crates/proposer-client/src/error.rs
```diff
@@ -28,4 +10,31 @@ pub enum Error {
...
+    Failed(String),
+
+    #[error("Grpc request error")]
+    Grpc(#[source] tonic::Status),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum GrpcConversionError {
+    #[error("Conversion of `{field}` failed")]
+    Conversion {
```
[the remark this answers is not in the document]
iljakuklic: > This looks like there'll only ever be a single variant, can we make it a simple struct?

The original idea here was that as the `.proto` file evolves, fields might be added and retired, so variants like `FieldMissing` etc might be introduced. Having said that, I don't have a strong opinion about converting it into a struct for now.

iljakuklic (dismissed):

crates/aggchain-proof-core/src/full_execution_proof.rs
```diff
@@ -12,6 +14,38 @@ use crate::{error::ProofError, vkey_hash::HashU32};
...
+    pub fn to_hash_u32(&self) -> HashU32 {
+        self.0
+            .into_iter()
+            .map(|n| n.as_canonical_u32())
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap()
+    }
+
+    pub fn to_hash_bn254(&self) -> [u8; 32] {
```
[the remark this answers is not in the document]
atanmarko: CC @hadjiszs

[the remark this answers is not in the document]
hadjiszs: to not block this PR I have propagated this here: https://github.com/agglayer/provers/issues/203

iljakuklic (dismissed):
LGTM. May be a good idea to wait for others to review the final version too, to get some more eyes on it.

iljakuklic (approved):

iljakuklic (dismissed):

crates/proposer-elfs/build.rs
```diff
@@ -1,12 +1,12 @@
...
     prover_elf_utils::ElfInfo::writing_to("vkeys_raw.rs")
         // Verification keys for aggregation proof
-        .module("aggregation", op_succinct_elfs::AGG_ELF)
+        .module("aggregation", op_succinct_elfs::AGGREGATION_ELF)
         .emit_vkey_bytes()
         .emit_vkey_hash()
         .finish()
         // Verification keys for range proof
-        .module("range", op_succinct_elfs::RANGE_ELF)
+        .module("range", op_succinct_elfs::RANGE_ELF_EMBEDDED)
```
Freyskeyd, line 9: why is it called `EMBEDDED` ?

crates/proposer-client/src/network_prover.rs
```diff
@@ -32,10 +32,10 @@ impl AggregationProver for NetworkProver {
     }
 }
 
-pub fn new_network_prover(endpoint: &str) -> anyhow::Result<NetworkProver> {
+pub fn new_network_prover(endpoint: &url::Url) -> anyhow::Result<NetworkProver> {
```
Freyskeyd, line 35: It could be just that (as Url is implementing `AsRef<str>`)
```suggestion
pub fn new_network_prover<T: AsRef<str>>(endpoint: T) -> anyhow::Result<NetworkProver> {
```

And just `rpc_url(endpoint)`

Freyskeyd (dismissed):
LGTM (with some nits)

crates/proposer-elfs/build.rs
atanmarko: CC @vcastellm

iljakuklic (approved):

Freyskeyd (approved):

PR #193, opened 2025-04-29
build: unify vkey crates

iljakuklic:
Context:

* separate crates no longer needed due to #183 
* separation introduced in #138
* builds on #191 
* obsoletes & takes patches from #179

## PR Checklist:


crates/proposer-elfs/Cargo.toml
```diff
@@ -6,10 +6,14 @@ license.workspace = true
 
 [dependencies]
 aggkit-prover-types = { workspace = true, features = ["sp1"] }
-proposer-vkeys-raw.workspace = true
 
 op-succinct-elfs.workspace = true
 
+[build-dependencies]
```
Ekleog-Polygon, line 12: Nit: can we have the build-deps above the deps? This way the order of reading the file is also the order in which deps are used

crates/prover-elf-utils/Cargo.toml
```diff
@@ -6,4 +6,4 @@ license.workspace = true
 
 [dependencies]
 bincode.workspace = true
-sp1-sdk.workspace = true
+sp1-prover = { workspace = true, features = [] }
```
Ekleog-Polygon, line 9:
```suggestion
sp1-prover.workspace = true
```

crates/prover-elf-utils/src/elf_info.rs
```diff
@@ -115,7 +115,8 @@ impl<ElfBytes: AsRef<[u8]>> Emitter<ElfBytes> {
 
     fn vkey(&mut self) -> &SP1VerifyingKey {
         self.vkey.get_or_insert_with(|| {
-            let (_pkey, vkey) = self.context.prover().setup(self.elf.as_ref());
+            let (_sp1_pkey, _stark_pkey, _program, vkey) =
+                self.context.prover().setup(self.elf.as_ref());
```
Ekleog-Polygon, line 119: I'm not sure I understand, why this change to SP1Prover with a different API?

Ekleog-Polygon (commented):
LGTM! Just a few comments below :)

crates/prover-elf-utils/src/elf_info.rs
iljakuklic: > I'm not sure I understand, why this change to SP1Prover with a different API?

Basically just to make the build dependencies lighter because I suspect this is on the build critical path. I don't know how much it actually helps but likely does not hurt.

atanmarko (approved):

atanmarko (dismissed):

atanmarko (approved):

PR #201, opened 2025-05-06
feat: add support for optimistic aggchain proof generation

Freyskeyd:
# Description

This PR is introducing the `optimistic-mode` for aggchain proof.

- Introduced `GenerateOptimisticAggchainProof` RPC in the gRPC service.
- Added `OptimisticAggchainProofInputs` struct for handling optimistic mode inputs.
- Updated `AggchainProofService` to handle both normal and optimistic proof requests.
- Refactored context generation logic into a reusable `Contextualize` trait.
- Updated protobuf definitions and generated code for the new RPC.

## PR Checklist:


crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -282,23 +291,29 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
+                    aggregation_proof_public_values
+                );
+
+                info!(
+                    "Aggregation proof public values retrieved from contracts: {:?}",
+                    retrieved_from_contracts
+                );
+
+                if aggregation_proof_public_values != &retrieved_from_contracts {
+                    return Err(Error::MismatchAggregationProofPublicValues {
```
atanmarko, line 312: On rust proposer branch, this `info` printouts are moved to error print within the if block. You can leave it this way, we will change it with the merge of the rust proposer branch.

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -61,15 +61,28 @@ pub(crate) type ProverService = Buffer<
 /// proof generation. Collected from various sources.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct AggchainProverInputs {
+    pub output_root: Digest,
```
atanmarko, line 64: Some comment explanation about the output_root origin/value meaning would be nice.

crates/aggkit-prover-types/src/error.rs
```diff
@@ -65,6 +65,20 @@ pub enum AggchainProofRequestError {
...
+        source: anyhow::Error,
+    },
+    #[error("Invalid AggchainProofRequest")]
+    InvalidAggchainProofRequest {
+        field_path: String,
+        source: anyhow::Error,
+    },
+    #[error("Missing AggchainProofRequest")]
+    MissingAggchainProofRequest { field_path: String },
+    #[error("Missing OptimisticMode signature")]
```
atanmarko, line 80: To be consistent use lower case for error strings.

crates/aggkit-prover/src/rpc.rs
```diff
@@ -73,70 +75,117 @@ impl AggchainProofGrpcService for GrpcService {
...
+                            proposer_client::Error::AggProofRequestFailed(
+                                jsonrpsee::core::client::error::Error::Call(e),
+                            ),
+                        ),
+                    ) if e.code() == -2000 => Err(Status::unavailable(
+                        "Proposer service has not built any proof yet",
+                    )),
+                    e => Err(Status::internal(e.to_string())),
+                }
+            }
```
atanmarko, line 141: Seems like fully duplicated code block for two kind of requests, should we move error processing to a helper function?

crates/aggkit-prover/src/utils/context.rs
```diff
@@ -0,0 +1,82 @@
+use std::collections::HashMap;
```
atanmarko, line 1: Should we move `Contextualize` trait to `aggchain-proof-types` and implementation to the relevant structure module (e.g. implementation for `AggchainProofInputs` to the `proof.rs`) ?

proto/aggkit/prover/v1/aggchain_proof_generation.proto
```diff
@@ -55,6 +65,18 @@ message GenerateAggchainProofResponse {
   bytes custom_chain_data = 5;
 }
 
+// The optimistic aggchain proof response message.
+message GenerateOptimisticAggchainProofResponse {
```
iljakuklic, line 69: I see the field numbers here match the normal aggchain proof. Out of curiosity, any particular reason for that?

crates/aggchain-proof-service/src/service.rs
```diff
@@ -209,3 +246,34 @@ impl tower::Service<AggchainProofServiceRequest> for AggchainProofService {
         .boxed()
     }
 }
+
+type AggchainProofServiceFuture =
+    Pin<Box<dyn Future<Output = Result<AggchainProofServiceResponse, Error>> + Send>>;
+
+impl tower::Service<AggchainProofServiceRequest> for AggchainProofService {
```
iljakuklic, line 253: Wold the following work? Instead of having one `tower::Service` impl for both modes with runtime dispatch, have two, `tower::Service<FullAggchainProofServiceRequest>` and `tower::Service<OptimisitcAggchainProofServiceRequest>`. That would allow to have a specific return type for each request type too.

crates/aggkit-prover/src/utils/context.rs
```diff
@@ -0,0 +1,82 @@
...
+    ($context: ident, $name:ident: $($data:tt)*) => {
+        $context.insert(stringify!($name).to_owned(), Bytes::from($($data)*.to_vec()));
+    };
+}
+
+macro_rules! int_to_bytes {
+    ($val:expr) => {
+        Bytes::from($val.to_be_bytes().to_vec())
+    };
+}
```
iljakuklic, line 17: I don't see this being used.

crates/aggkit-prover/src/utils/context.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use aggchain_proof_types::{AggchainProofInputs, OptimisticAggchainProofInputs};
+use agglayer_interop::types::U256;
+use prost::bytes::Bytes;
+
+macro_rules! context_field {
+    ($context: ident, $name:ident: $($data:tt)*) => {
+        $context.insert(stringify!($name).to_owned(), Bytes::from($($data)*.to_vec()));
+    };
+}
```
iljakuklic, line 11: Style nitpick:

Maybe the macro could use repetition to make it look a bit more like a struct constructor.

I am thinking of approximately the following syntax at use sites:
```rust
let context = context! {
    last_proven_block: self.last_proven_block.to_be_bytes(),
    requested_end_block: self.requested_end_block.to_be_bytes(),
    l1_info_tree_root_hash: self.l1_info_tree_root_hash.as_bytes(),
    l1_info_tree_index: self.l1_info_tree_leaf.l1_info_tree_index.to_be_bytes(),
    l1_info_tree_rer: self.l1_info_tree_leaf.rer.as_bytes(),
    l1_info_tree_mer: self.l1_info_tree_leaf.mer.as_bytes(),
    l1_info_tree_ger: self.l1_info_tree_leaf.inner.global_exit_root.as_bytes(),
    l1_info_tree_block_hash: self.l1_info_tree_leaf.inner.block_hash.as_bytes(),
    l1_info_tree_timestamp: self.l1_info_tree_leaf.inner.timestamp.to_be_bytes(),
};
```

EDIT: this code just moved here. Feel free to ignore.

iljakuklic (commented):
Looks good to me overall. Some comments/suggestions below.

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -61,15 +61,28 @@ pub(crate) type ProverService = Buffer<
...
     pub stdin: SP1Stdin,
-    pub last_proven_block: u64,
-    pub end_block: u64,
+}
+
+pub enum AggchainProofMode {
```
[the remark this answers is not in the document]
Freyskeyd: Yeah I though about the same, maybe we can rename this to `FepVerification` ?
cc @hadjiszs do you have any input?

crates/aggchain-proof-core/src/full_execution_proof.rs
```diff
@@ -37,7 +37,7 @@ pub struct FepInputs {
     /// Trusted sequencer address.
     pub trusted_sequencer: Address,
     /// Signature in the "OptimisticMode" case.
-    pub signature_optimistic_mode: Option<PrimitiveSignature>,
+    pub signature_optimistic_mode: Option<agglayer_primitives::Signature>,
```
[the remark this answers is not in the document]
Freyskeyd: I think it is fine to use this one 👍🏼

crates/aggchain-proof-service/src/service.rs
Freyskeyd: But the return type for both "mode" is the same tho

crates/aggchain-proof-builder/src/lib.rs
hadjiszs: I think it's very valuable to keep printing them in all cases, even if no mismatch here, because there is still room for mismatch with the L1 much later in the flow via the `aggchain_hash`

crates/aggchain-proof-service/src/service.rs
iljakuklic: > But it means that you'll need to have multiple instance of this service on the upper function.

Ok, in that case, my suggestion does not make that much sense.

crates/aggchain-proof-builder/src/lib.rs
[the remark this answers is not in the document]
hadjiszs: maybe `pub enum OptimisticMode` with the two variants `Enabled { .. }` and `Disabled { .. }` ?

crates/aggkit-prover/src/utils/context.rs
Freyskeyd: I don't think it is valuable to move this to `aggchain-proof-types` as the `Contextualize` trait is only used for providing `gRPC`'s context for the query. A better place could be in `aggkit-prover-types` as the `Contextualize` trait is using prost's `Bytes`.

crates/aggkit-prover/src/utils/context.rs
Freyskeyd: It is used in the `ger_leaves`

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -282,23 +291,29 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
-
-            if request.aggregation_proof_public_values != retrieved_from_contracts {
-                return Err(Error::MismatchAggregationProofPublicValues {
-                    expected_by_contract: Box::new(retrieved_from_contracts),
-                    expected_by_verifier: Box::new(request.aggregation_proof_public_values),
-                });
+            if let AggchainProofMode::Proof {
+                ref aggregation_proof_public_values,
+                ..
+            } = request.aggchain_proof_mode
```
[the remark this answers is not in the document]
Freyskeyd: But what would contain the other branches?

crates/aggchain-proof-builder/src/lib.rs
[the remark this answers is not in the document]
Freyskeyd: I changed this to `FepVerification` 

`OptimisticModeWithPayload` feels like it's only for optimistic mode

crates/aggchain-proof-builder/src/lib.rs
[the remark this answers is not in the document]
iljakuklic: > But what would contain the other branches?

Just plug it up with the unit type, something like:
```rust
match {
    AggchainProofMode::Proof { /* omitted */ } => {
        the_current_if_branch_body();
    }
    AggchainProofMode::Optimistic { .. } => (),
}
```

crates/aggchain-proof-builder/src/lib.rs
Freyskeyd: Added as part of the doc comment for `OutputRoot`

crates/aggkit-prover/src/rpc.rs
Freyskeyd: That can be handled when we properly define the aggkit prover gRPC errors

crates/aggchain-proof-builder/src/lib.rs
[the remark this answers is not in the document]
Freyskeyd: I'm not quite sure if it's valuable to add this, if we're adding a new "mode", which is unlikely to happen, we'll for sure refactor this part and then switch to a match if needed.

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -281,68 +290,82 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
-                    request.aggregation_proof_public_values
-                );
-                return Err(Error::MismatchAggregationProofPublicValues {
-                    expected_by_contract: Box::new(retrieved_from_contracts),
-                    expected_by_verifier: Box::new(request.aggregation_proof_public_values),
-                });
+            if let FepVerification::Proof {
+                ref aggregation_proof_public_values,
+                ..
+            } = request.fep_verification
```
hadjiszs, line 296: else if `FepVerification::Optimistic { signature }` then to set `fep_inputs.signature_optimistic_mode = Some(signature)`

I think this could be plugged in your PR right?

crates/aggchain-proof-core/src/full_execution_proof.rs
```diff
@@ -240,16 +271,17 @@ impl FepInputs {
...
             self.prev_state_root.0,
             self.prev_withdrawal_storage_root.0,
             self.prev_block_hash.0,
         )
+        .into()
     }
 
     /// Compute claim root.
-    pub fn compute_claim_root(&self) -> Digest {
+    pub fn compute_claim_root(&self) -> OutputRoot {
```
hadjiszs, line 284: nit: I think `L2PreRoot` is also an `OutputRoot`. If we want to bring distinct type to help us differentiate between `l2_pre_root` and `claim_root` (aka. `l2_post_root`). Then I would use `L2PreRoot` and `ClaimRoot` instead

```suggestion
    pub fn compute_claim_root(&self) -> ClaimRoot {
```

(or `L2PostRoot` but then need rename the functions and other `claim_root` occurrences)

crates/aggchain-proof-core/src/full_execution_proof.rs
```diff
@@ -1,6 +1,6 @@
 use agglayer_primitives::keccak::keccak256_combine;
```
hadjiszs, line 1: I think need the aggchain proof elf to be updated

crates/aggchain-proof-core/src/full_execution_proof.rs
```diff
@@ -14,6 +14,37 @@ use crate::{error::ProofError, vkey_hash::HashU32};
...
+}
+
+/// OutputRoot is the concatenation of the OutputRoot version + payload
+///
+/// Payload composed of `state_root`, `withdrawal_storage_root`,
+/// `latest_block_hash`
+///
+/// Details: https://specs.optimism.io/protocol/proposals.html#l2-output-commitment-construction
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct OutputRoot(pub Digest);
```
iljakuklic, line 34: Nit: Some conversion methods like `to_array`, `to_fixed_bytes`, etc. with explicit names would be nice to avoid the `.0 .0` field accesses.

crates/aggchain-proof-core/src/full_execution_proof.rs
```diff
@@ -14,6 +14,37 @@ use crate::{error::ProofError, vkey_hash::HashU32};
...
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct L2PreRoot(pub Digest);
+
+impl From<L2PreRoot> for FixedBytes<32> {
+    fn from(value: L2PreRoot) -> FixedBytes<32> {
+        value.0 .0.into()
+    }
+}
+
+/// OutputRoot is the concatenation of the OutputRoot version + payload
```
iljakuklic, line 27:
```suggestion
/// OutputRoot is the hash of the concatenation of the OutputRoot version + payload
```

crates/aggchain-proof-core/src/full_execution_proof.rs
Freyskeyd: I wasn't sure about the naming for those 2, I'm fine switching to `ClaimRoot`

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -290,80 +290,82 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
+
+                info!("Aggchain-params unrolled values: {aggchain_params_values:?}");
+                info!(
+                    "Aggchain-params abi encoded packed: {}",
+                    hex::encode(fep_inputs.encoded_aggchain_params())
+                );
+                info!(
+                    "Aggchain-params keccak-hashed: {}",
+                    fep_inputs.aggchain_params()
+                );
```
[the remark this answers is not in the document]
Freyskeyd: Those logs are from another PRs, @hadjiszs I think that's you decision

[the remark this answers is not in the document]
hadjiszs: I would definitely keep the first log as `info`, but the 2 after can be removed or put as `debug`

atanmarko (approved):

iljakuklic (approved):

PR #204, opened 2025-05-12
feat: proposer client-service tests

### LINKED ISSUE #96: Implement missing `proposer-client` tests

atanmarko:
# Description

Additional proposer client tests.

Fixes #96


crates/proposer-client/src/tests/proposer_rpc.rs
```diff
@@ -12,85 +13,215 @@ use crate::{
...
+        + Send
+        + Sync
+        + 'static,
+{
+    let mut server = MockProofsService::new();
+
+    server
+        .expect_request_agg_proof()
+        .with(mockall::predicate::function(
+            move |req: &tonic::Request<AggProofRequest>| req.get_ref() == &request.clone(),
```
Ekleog-Polygon, line 42: `&request.clone()` looks quite weird, what pushed you towards this? I'd expect just `&request` to work fine here; and maybe even just using the `.eq` filter instead of `.with`, at which point the function might not even be needed any longer.

crates/proposer-client/src/tests/proposer_rpc.rs
```diff
@@ -12,85 +13,215 @@ use crate::{
...
-#[test]
-#[ignore = "to be implemented"]
-fn request_with_an_invalid_request_id() {}
+    let service = ProposerRpcClient::new(server.uri(), Duration::from_millis(500))
+        .await
+        .unwrap();
+    let response = service.request_agg_proof(request.clone()).await;
+
+    assert!(response.is_err());
+    if let Err(crate::error::Error::Requesting(crate::error::ProofRequestError::Grpc(status))) =
```
Ekleog-Polygon, line 122: Nit: use `response.unwrap_err` if there's no too cumbersome `Debug` bound

crates/proposer-client/src/tests/proposer_rpc.rs
```diff
@@ -12,85 +13,215 @@ use crate::{
...
+        response
+    {
+        assert_eq!(status.code(), tonic::Code::InvalidArgument);
+        assert_eq!(
+            status.message(),
+            "Requested end block must be greater than the last proven block"
```
Ekleog-Polygon, line 220: Do we really want to assert the exact error messages here? If yes, can we use `insta` so it's easier to update whenever we want to change the error messages?

Ekleog-Polygon (commented):
LGTM! I just have a few comments down below 😄

iljakuklic (approved):

Ekleog-Polygon (approved):

PR #221, opened 2025-05-23
chore: re-add shutdown timeout for prover engine

hadjiszs:
# Description

Shutdown timeout safeguard in case that the cancellation token haven't completed on time.

## PR Checklist:

### DIFF
--- a/crates/prover-engine/src/lib.rs
+++ b/crates/prover-engine/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{convert::Infallible, future::IntoFuture, net::SocketAddr};
+use std::{convert::Infallible, future::IntoFuture, net::SocketAddr, time::Duration};
 
 use agglayer_telemetry::ServerBuilder as MetricsBuilder;
 use http::{Request, Response};
@@ -22,6 +22,7 @@ pub struct ProverEngine {
     cancellation_token: Option<CancellationToken>,
     metric_socket_addr: Option<SocketAddr>,
     rpc_socket_addr: Option<SocketAddr>,
+    shutdown_runtime_timeout: Option<Duration>,
 }
 
 impl ProverEngine {
@@ -35,6 +36,7 @@ impl ProverEngine {
             cancellation_token: None,
             metric_socket_addr: None,
             rpc_socket_addr: None,
+            shutdown_runtime_timeout: None,
         }
     }
 
@@ -72,6 +74,13 @@ impl ProverEngine {
 
         self
     }
+
+    pub fn set_shutdown_runtime_timeout(mut self, timeout: Duration) -> Self {
+        self.shutdown_runtime_timeout = Some(timeout);
+
+        self
+    }
+
     pub fn add_rpc_service<S>(mut self, rpc_service: S) -> Self
     where
         S: Service<Request<BoxBody>, Response = Response<BoxBody>, Error = Infallible>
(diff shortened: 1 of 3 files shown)


crates/prover-engine/src/lib.rs
```diff
@@ -22,6 +22,7 @@ pub struct ProverEngine {
     cancellation_token: Option<CancellationToken>,
     metric_socket_addr: Option<SocketAddr>,
     rpc_socket_addr: Option<SocketAddr>,
+    shutdown_runtime_timeout: Option<Duration>,
```
iljakuklic, line 25: Nit: the following feels a bit more natural to me:

```suggestion
    runtime_shutdown_timeout: Option<Duration>,
```

crates/prover-engine/src/lib.rs
```diff
@@ -72,6 +74,13 @@ impl ProverEngine {
 
         self
     }
+
+    pub fn set_shutdown_runtime_timeout(mut self, timeout: Duration) -> Self {
```
iljakuklic, line 78: Also here

iljakuklic (approved):

crates/prover-engine/src/lib.rs
```diff
@@ -235,8 +244,10 @@ impl ProverEngine {
                 }
             });
 
-        // prover_runtime.shutdown_timeout(config.shutdown.runtime_timeout);
-        // metrics_runtime.shutdown_timeout(config.shutdown.runtime_timeout);
+        if let Some(timeout) = self.runtime_shutdown_timeout {
```
Freyskeyd, line 247: Maybe instead of having it optional, we could fallback to a constant duration, I think 30s is the standard in terms of signal shutdown

Ekleog-Polygon: What I can say, from my ongoing work on the certificate task, is that we're currently waiting for each inflight certificate to finish processing before cancelling its network task.

I do think that, with the way things are currently written around there (and all the doubts around unclean-shutdown-related possible database corruption), 30s is too short.

Once my work on certificate tasks finishes, as well as the follow-up work on cleaning everything up, we can probably reconsider... but for now I'd see anything below ~1hr to be a risky timeout.

Does that make sense? And if so, can we clearly document it in the option text? (I'd personally feel even better not having the shutdown timeout at all until that work completes, but I guess we're doing this PR for a reason)

Freyskeyd: > What I can say, from my ongoing work on the certificate task, is that we're currently waiting for each inflight certificate to finish processing before cancelling its network task.
> 
> I do think that, with the way things are currently written around there (and all the doubts around unclean-shutdown-related possible database corruption), 30s is too short.
> 
> Once my work on certificate tasks finishes, as well as the follow-up work on cleaning everything up, we can probably reconsider... but for now I'd see anything below ~1hr to be a risky timeout.
> 
> Does that make sense? And if so, can we clearly document it in the option text? (I'd personally feel even better not having the shutdown timeout at all until that work completes, but I guess we're doing this PR for a reason)

The problem is that we do not have control over that.
When node gets reallocated (moved to another instance), the container receives a SIGTERM signal, then the orchestrator (the system one) is awaiting 30s, then send a SIGKILL and kill the process/container.

Ekleog-Polygon: Oh. Then such a timeout is basically already enforced anyway 🥲

Well, I guess landing this won't make things worse at least, and I hope to be able to do that work revamping the certificate task soon!

Ekleog-Polygon (approved):

PR #228, opened 2025-06-02
feat: add otlp tracing

### LINKED ISSUE #227: Add alternative open telemetry output for traces

atanmarko:
Add open telemetry output as possible output for traces.

Fixes #227


crates/prover-tracer/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+            let mut v = tag_raw.splitn(2, '=');
+            match (v.next(), v.next()) {
+                (Some(key), Some(value)) if !key.trim().is_empty() && !value.trim().is_empty() => {
+                    Some(KeyValue::new(
+                        key.trim().to_string(),
+                        value.trim().to_string(),
+                    ))
+                }
+                _ => None,
+            }
```
iljakuklic, line 276: I think you could use `split_once` here:
```suggestion
            let (key, val) = tag_raw.split_once(2, '=')?;
            let key = key.trim();
            let val = val.trim();
            if key.is_empty() || val.is_empty() {
                return None;
            }
            Some((key.to_string(), val.to_string()))
```

crates/prover-tracer/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+        };
+
+        write!(f, "{level}")
+    }
+}
+
+impl From<TracingLevel> for EnvFilter {
+    fn from(value: TracingLevel) -> Self {
+        EnvFilter::new(format!(
+            "warn,prover={value},aggkit={value},agglayer={value},pessimistic_proof={value}"
```
iljakuklic, line 77: Where are the `prover=`, `aggkit=` etc. keys coming from? I assumed these are supposed to be full crate names but might be wrong.

crates/prover-tracer/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+/// appropriate enum variant. If the string is not recognized to be either
+/// `stdout` or `stderr`, it is assumed to be a file path.
+#[derive(Serialize, Debug, Clone, Default, PartialEq, Eq)]
+pub enum TracingOutput {
+    #[default]
+    Stdout,
+    Stderr,
+    File(PathBuf),
+    Otlp,
+}
```
iljakuklic, line 97: Could something like this be used to avoid the custom deserialize below?
```suggestion
#[derive(Serialize, Debug, Clone, Default, PartialEq, Eq, Deserialize)]
pub enum TracingOutput {
    #[default]
    Stdout,
    Stderr,
    Otlp,
    #[serde(untagged)]
    File(PathBuf),
}
```

crates/prover-tracer/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+                .with_max_queue_size(match std::env::var("OTLP_BATCH_MAX_QUEUE_SIZE") {
+                    Ok(v) => v.parse::<usize>().unwrap_or(2048),
+                    _ => 2048,
+                })
+                .with_max_export_batch_size(
+                    match std::env::var("OTLP_BATCH_MAX_EXPORTER_BATCH_SIZE") {
+                        Ok(v) => v.parse::<usize>().unwrap_or(512),
+                        _ => 512,
+                    },
+                );
```
iljakuklic, line 165: Here too, the various `parse` calls could return an error rather than silently accept a malformed input and fabricate a value out of thin air. Same further down below this comment.

iljakuklic (commented):
A bunch of nits:

PR #236, opened 2025-06-11
chore: update proposer to v2.3.1

atanmarko:


crates/aggchain-proof-program/elf/riscv32im-succinct-zkvm-elf
iljakuklic, line 1: Do we know why this needed a rebuild?

iljakuklic (approved):

PR #237, opened 2025-06-12
ci: elf checking fixes

iljakuklic:
* Make the elf checking run on the `push` event.
* Use a fixed version for `cargo prove`.

## PR Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,7 +65,7 @@ jobs:
 
   check_elf:
     runs-on: ubuntu-latest-16-cores
-    if: ${{ contains('["merge_group", "workflow_dispatch"]', github.event_name) }}
+    if: ${{ contains('["push", "workflow_dispatch"]', github.event_name) }}
     name: ELF Checking
     steps:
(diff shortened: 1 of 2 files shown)


atanmarko (approved):

.github/workflows/test.yml
```diff
@@ -65,7 +65,7 @@ jobs:
 
   check_elf:
     runs-on: ubuntu-latest-16-cores
-    if: ${{ contains('["merge_group", "workflow_dispatch"]', github.event_name) }}
+    if: ${{ contains('["push", "workflow_dispatch"]', github.event_name) }}
```
Ekleog-Polygon, line 68: Hmm I think we still need merge_group checks? Otherwise we could end up with an ill-synchronized ELF binary, eg. if one PR adds a call to the foo() function, and before it lands another PR lands that modifies the contents of the foo function.

Ekleog-Polygon (changes requested):
I'm happy with the fixed version; but the CI changes look wrong to me. Did I miss something?

PR #239, opened 2025-06-12
feat: make static call caller address as config

hadjiszs:
# Description

Workaround attempt for the nonce check issue:

- `static_call_caller_address` is now an optional config parameter to avoid the conflicting zero address
- The default address is `0x39027D57969aD59161365e0bbd53D2F63eE5AAA6` if not provided

## PR Checklist:


crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -139,6 +139,9 @@ pub struct AggchainProofBuilder<ContractsClient> {
 
     /// Verification key for the aggchain proof.
     aggchain_vkey: Arc<SP1VerifyingKey>,
+
+    /// Static call caller address.
+    static_call_caller_address: Address,
```
atanmarko, line 144: @hadjiszs Do we need this here, or we can keep it just in the `ContractsClient`?

atanmarko: EDIT: I see it is used in the bridge witness. Not to change API of the `ContractsClient` with getter, better to leave it here.

atanmarko (approved):

atanmarko (approved):

PR #262, opened 2025-08-04
fix: return differentiated errors from InsertedGER::verify instead of bool

VolodymyrBg:
Replace boolean return type with Result<(), GerVerificationError> in InsertedGER::verify

- Add GerVerificationError enum with specific error variants:
  - L1InfoRootMismatch: when L1 info root doesn't match proof root
  - MerkleProofVerificationFailed: when Merkle proof verification fails
- Update verify function to return detailed error information
- Remove TODO comment as differentiated errors are now implemented
- Improve code readability by extracting leaf_hash and index variables
- Add comprehensive documentation for error types
- Fix usage in bridge/mod.rs to handle new Result return type

This change provides better error handling and debugging capabilities
by giving specific information about what went wrong during GER verification.
The function now returns meaningful error messages instead of just true/false,
making it easier to diagnose issues in production environments.

### DIFF
--- a/crates/aggchain-proof-core/src/bridge/mod.rs
+++ b/crates/aggchain-proof-core/src/bridge/mod.rs
@@ -384,7 +384,7 @@ impl BridgeConstraintsInput {
             .bridge_witness
             .inserted_gers
             .iter()
-            .find(|ger| !ger.verify(self.l1_info_root));
+            .find(|ger| ger.verify(self.l1_info_root).is_err());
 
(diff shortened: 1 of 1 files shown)


Ekleog-Polygon (commented):
Conceptually this LGTM! Unfortunately we can't accept external contributions yet. We'll come back to you once the process for external contributors is defined.

Also, this is more a matter of implementation, but IMO `GerVerificationError`, considering it's only used at a single place for now, would deserve being a ZST like `pub struct GerVerificationError;` :)

Thank you for your PR anyway!

Issue #206, opened 2025-05-13
aggkit-prover exits silently if already used port is set for GRPC


xavier-romero:
Setting this config:
`grpc-endpoint = "0.0.0.0:4446"`

To a port that's already used on the machine, makes the program exit silently with no error, warn or anything.
A proper log would be useful.

PR #264, opened 2025-08-04
fix: exit with error

atanmarko:
Exit with error in `aggkit-prover` and `agglayer-prover`.

Fixes [# (issue)](https://github.com/agglayer/provers/issues/206)


iljakuklic (commented):
There is a Github warning:

> The head ref may contain hidden characters: "fix/prover-\u0435xit-with-error"

Looks good otherwise.

PR #269, opened 2025-08-07
build: adding merge-group

Freyskeyd:
This PR adds the support of merge group


polykeith (dismissed):

iljakuklic (approved):
Not sure if this is sufficient but let's give it a go.

PR #271, opened 2025-08-11
WIP feat: remove gers and unclaims

atanmarko:
Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)


crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -279,6 +280,28 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
+        let removed_gers: Vec<Digest> = request
+            .aggchain_proof_inputs
+            .removed_gers
+            .into_iter()
+            .filter(|removed_ger| new_blocks_range.contains(&removed_ger.block_number))
+            .collect::<Vec<_>>()
+            .into_iter()
+            .sorted()
+            .map(|removed_ger| removed_ger.global_exit_root)
+            .collect();
```
polykeith, line 292:
```suggestion
        let mut removed_gers = request
            .aggchain_proof_inputs
            .removed_gers
            .into_iter()
            .filter(|removed_ger| new_blocks_range.contains(&removed_ger.block_number))
            .collect::<Vec<_>>();
        removed_gers.sort();
        let removed_gers: Vec<Digest> = removed_gers
            .into_iter()
            .map(|removed_ger| removed_ger.global_exit_root)
            .collect();
```

That way, we can eliminate the need for `itertools` altogether.

Ekleog-Polygon (dismissed):
LGTM! :smile:

crates/aggchain-proof-core/src/bridge/mod.rs
```diff
@@ -328,16 +321,17 @@ impl BridgeConstraintsInput {
...
-        // global indices.
         let constrained_claims = ImportedBridgeExitCommitmentValues {
             claims: filter_values(
-                &self.bridge_witness.global_indices_unset,
-                &self.bridge_witness.bridge_exits_claimed,
-                |exit: &GlobalIndexWithLeafHash| -> U256 { exit.global_index },
+                &self.bridge_witness.unset_claims, // Vec<Digest> of unclaimed hashes
+                &self.bridge_witness.bridge_exits_claimed, // Vec<GlobalIndexWithLeafHash>
+                |exit: &GlobalIndexWithLeafHash| -> Digest {
+                    exit.bridge_exit_hash // compare by hash
```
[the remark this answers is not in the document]
invocamanman: totally!!

crates/aggchain-proof-types/src/inserted_ger.rs
```diff
@@ -11,7 +13,28 @@ pub struct InsertedGerWithBlockNumber {
...
+impl Ord for InsertedGerWithBlockNumber {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.block_number.cmp(&other.block_number).then_with(|| {
+            let ordering = self.block_index.cmp(&other.block_index);
+            // Debug assert that if block_number and block_index are equal,
+            // then inserted_ger should also be equal to maintain Ord guarantees.
+            debug_assert!(
+                ordering != Ordering::Equal || self.inserted_ger == other.inserted_ger,
+                "Items with same block_number and block_index must have same inserted_ger"
+            );
```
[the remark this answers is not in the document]
atanmarko: That is indeed good question if we want to panic in production here - thanks for the feedback!

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -267,17 +267,55 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
+        // Prepare removed GERS for the proof.
+        let removed_gers: Vec<Digest> = filter_sort_map(
+            request.aggchain_proof_inputs.removed_gers,
+            &new_blocks_range,
+            |removed_ger| removed_ger.block_number,
+            |removed_ger| removed_ger.global_exit_root,
+        );
+
+        // Prepare unset claims for the proof.
+        let unset_claims: Vec<Digest> = filter_sort_map(
```
invocamanman, line 313: I think we have an error here, as @hadjiszs pointed out before, something wasn't ok.

  ```
unsetGlobalIndexHashChain = Hashes.efficientKeccak256(
                unsetGlobalIndexHashChain,
                bytes32(globalIndex)
            );
```
The `unset_claims` are a vector of `globalIndex` indicating uniquely which leafs were unclaimed
```
     * @param globalIndexes Global index is defined as:
     * | 191 bits |    1 bit     |   32 bits   |     32 bits    |
     * |    0     |  mainnetFlag | rollupIndex | localRootIndex |
```

¡

File: crates/aggchain-proof-builder/src/lib.rs [hunk 5]
```diff
@@ -261,27 +261,80 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
             .await
             .map_err(Error::UnableToFetchTrustedSequencerAddress)?;
 
-        // From the request
-        let inserted_gers: Vec<InsertedGER> = request
+        // Helper function to filter, sort, and convert a vector of items to the desired
+        // output type. It filters items based on a range of blocks, sorts them
+        // using type Ord implementation, and maps them to a new type using the
+        // provided mapping function.
+        fn filter_sort_map<T, F, U>(
```
iljakuklic, line 268: Nit: In order to make the method more generic, it could take `items` as `impl IntoIterator<Item = T>` and return an `impl Iterator<Item = U>`. That would avoid a number of intermediate `Vec`s introduced below.

crates/aggchain-proof-core/src/bridge/mod.rs
```diff
@@ -328,15 +328,13 @@ impl BridgeConstraintsInput {
...
-        // global indices.
+    /// Verify that the claims filtered with the global indexes that got
+    /// unclaimed are equal to commited imported bridge exits.
+    pub fn verify_constrained_claims(&self) -> Result<(), BridgeConstraintsError> {
         let constrained_claims = ImportedBridgeExitCommitmentValues {
             claims: filter_values(
-                &self.bridge_witness.global_indices_unset,
-                &self.bridge_witness.bridge_exits_claimed,
+                &self.bridge_witness.unset_claims, // Vec<U256> of unclaimed global indexes
+                &self.bridge_witness.bridge_exits_claimed, // Vec<GlobalIndexWithLeafHash>
```
iljakuklic, line 337: Nit: This seems like a fairly useless comment
```suggestion
                &self.bridge_witness.bridge_exits_claimed,
```

crates/aggchain-proof-types/src/imported_bridge_exit.rs
```diff
@@ -1,14 +1,37 @@
+use std::cmp::Ordering;
```
iljakuklic, line 1: Nit: fully qualified `std::cmp::Ordering` used twice below.

iljakuklic (commented):
A bunch of nits:

crates/aggchain-proof-types/src/inserted_ger.rs
```diff
@@ -11,7 +13,28 @@ pub struct InsertedGerWithBlockNumber {
...
+impl Ord for InsertedGerWithBlockNumber {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.block_number.cmp(&other.block_number).then_with(|| {
+            let ordering = self.block_index.cmp(&other.block_index);
+            // Assert that if block_number and block_index are equal,
+            // then inserted_ger should also be equal to maintain Ord guarantees.
+            assert!(
+                ordering != Ordering::Equal || self.inserted_ger == other.inserted_ger,
+                "Items with same block_number and block_index must have same inserted_ger"
+            );
```
Ekleog-Polygon: Same here

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -537,7 +537,7 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
                 .commitment(IMPORTED_BRIDGE_EXIT_COMMITMENT_VERSION),
                 bridge_witness: BridgeWitness {
                     inserted_gers,
-                    bridge_exits_claimed: all_imported_bridge_exits,
+                    claimed_imported_bridge_exits: all_imported_bridge_exits,
```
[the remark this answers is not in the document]
atanmarko: @Ekleog-Polygon `all_imported_bridge_exits` ==  All the bridge exits in the new blocks range, also those that are unclaimed. It means, I want to emphasize the fact that unclaimed bridge exits are part of this array.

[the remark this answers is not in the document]
atanmarko: Having said that, I should also rename `bridge_exits_claimed`/`claimed_imported_bridge_exits` to `all_imported_bridge_exits` (or just `imported_bridge_exits`)

crates/aggchain-proof-types/src/imported_bridge_exit.rs
```diff
@@ -1,14 +1,37 @@
...
+impl Ord for ImportedBridgeExitWithBlockNumber {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.block_number.cmp(&other.block_number).then_with(|| {
+            let ordering = self.global_index.cmp(&other.global_index);
+            // Assert that if block_number and global_index are equal,
+            // then bridge_exit_hash should also be equal to maintain Ord guarantees.
+            assert!(
+                ordering != Ordering::Equal || self.bridge_exit_hash == other.bridge_exit_hash,
+                "Items with same block_number and global_index must have the same bridge_exit_hash"
+            );
```
invocamanman: is this assumptions always true?¿, why we order them by global index, and not by transaction index?
Also coudl happen in case there's an unset claim, that same global index is claimed but have different or the same hash

invocamanman: also just check it
all the rest we sort by block_index,
crates/aggchain-proof-types/src/imported_bridge_exit.rs it's the only one htat we use global_index and we shouldn't

Finally...
if a single transaction make 2 updates.... that would be hard to check... i guess the only way would be to retrieve the tx hash and check the order of the events... but better to reorder in a "random" way i would respect the order were this is sent... i guess is better than nothing

invocamanman: @atanmarko

invocamanman:
As per me i think all the comparators has an "extra" criteria, after the logIndex.
The logIndex inside a blocknumber is unique so a case were it matches should at least generate a warning IMO.
The final criteria like: `.then_with(|| self.bridge_exit_hash.cmp(&other.bridge_exit_hash))` it's arbitrary, + it could happen that 2 bridges has the same bridgeHash, so it's not unique either.

invocamanman:
If a requirement of ord is that there cannot be equal elements, it could happen in the current criteria :thinking: 
Plus in the 20% if we order them "randomly" we have a 50% chance of not following correctly the hashChain, therefore being unable to generate a proof D:. I think we cannot do better than that anyway xD but i would warn that should be unreachable code and add a comment from audit perspective ^^

crates/aggchain-proof-core/src/bridge/mod.rs
```diff
@@ -328,15 +328,13 @@ impl BridgeConstraintsInput {
         Ok(bridge_address.into())
     }
 
-    /// Verify that the claimed global indexes minus the unset global indexes
-    /// are equal to the Constrained global indexes.
-    fn verify_constrained_global_indices(&self) -> Result<(), BridgeConstraintsError> {
-        // Check if the filtered claimed global indices are equal to the constrained
-        // global indices.
+    /// Verify that the claims filtered with the global indexes that got
+    /// unclaimed are equal to commited imported bridge exits.
```
hadjiszs:
```suggestion
    /// unclaimed are equal to committed imported bridge exits.
```

PR #296, opened 2025-09-18
feat: rework succinct contract config

atanmarko:
Fetch op-succinct contract config from the contract, not from the local configuration.


atanmarko:
Wrong branch for merge, rebasing

PR #304, opened 2025-10-02
fix: grpc max decoding/encoding size

atanmarko:
Fix grpc config.
 
Switching to new tonic version, communication was breaking due to limited message size.


Ekleog-Polygon (approved):
IIUC the failure was silent and the PR that introduced it didn't trigger any test; do you know how hard it'd be to add a regression test that'd expose the bug earlier?

iljakuklic (approved):

PR #313, opened 2025-10-27
feat: moving agglayer-prover to agglayer repo

Freyskeyd:
BREAKING-CHANGE: `agglayer-prover` is moved to `agglayer` repo, gRPC import will not work.


Ekleog-Polygon (approved):
LGTM, though we should probably make sure that we can undraft and land https://github.com/agglayer/agglayer/pull/1141 ASAP so that we don't block all development on agglayer-prover in the meantime :smile:

PR #314, opened 2025-10-28
chore: bump op-succinct-elfs to v3.3.2 and update tests

vcastellm:
This pull request updates the dependency version for `op-succinct-elfs` and refreshes snapshot test data for aggregation and range verification keys. The main focus is ensuring compatibility with the latest upstream changes and keeping test outputs in sync.

### DIFF
--- a/crates/prover-logger/src/log.rs
+++ b/crates/prover-logger/src/log.rs
@@ -50,7 +50,10 @@ impl Display for LogLevel {
 impl From<LogLevel> for EnvFilter {
     fn from(value: LogLevel) -> Self {
         EnvFilter::new(format!(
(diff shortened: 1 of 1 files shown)


atanmarko:
>  I just have one question as I have less experience with the aggchain-proof: do we have to bump a version like we do for PP selector whenever we modify the AP program? I'd guess no considering this is already passing tests, but need to be sure before r+'ing here.

@Ekleog-Polygon Yes when we change aggchain program - no when we change op-succinct dependencies, as range and aggregation vkeys are passed as parameter to aggchain program

Ekleog-Polygon (dismissed):
LGTM, thanks!

atanmarko (approved):

atanmarko (approved):

PR #345, opened 2026-04-17
feat: range reduction breaks paired claims / unclaims

### LINKED ISSUE #343: bug: range reduction removes an unclaim that is paired with a claim still present in the certificate
## Description 

When a `GenerateAggchainProof` request is built for a block range [`last_proven_block`, `requested_end_block`], both claims (imported_bridge_exits) and unclaims (unclaims) are collected for all events that fall within that range.

The problem arises when the block range is reduced:

A bridge exit was claimed at block X, where X ≤ end_block' → the claim remains in imported_bridge_exits.
That same claim was later unclaimed at block Y, where end_block' < Y ≤ original_end_block → the unclaim is now outside the new range and gets dropped from unclaims.
The result is an inconsistent certificate: it contains an imported_bridge_exit (claim) with no corresponding unclaim, even though on-chain the claim was reversed. The proof circuit requires that claim/unclaim pairs are coherent within the block range — a claim whose unclaim falls outside the range cannot be safely included, because the verifier would see a net claimed bridge exit that was actually cancelled.

In short: reducing the block range can silently split a claim/unclaim pair across the range boundary, leaving an orphaned claim in the certificate whose cancellation is invisible to the proof. This makes the proof either unprovable or incorrect with respect to the actual on-chain state.

## Proposed solution: early rejection with a suggested cutoff hint

Instead of proceeding with an expensive proof generation for a block range known to be invalid, `aggkit-prover` will detect the inconsistency upfront — specifically, that reducing the range to `end_block` would drop an unclaim whose paired claim is still included — and immediately return an error to the caller.

The error includes the problematic cutoff block (`end_block`) as context, signalling to aggsender that this particular range boundary is unsafe. `aggsender` can then search for a new cutoff point strictly below `end_block` — one that either includes both the claim and its unclaim, or excludes both — and resubmit the GenerateAggchainProof request with the corrected range.

This avoids the cost of running a ZK proof (which involves SP1 network proving time and resources) only to produce an invalid or unprovable certificate. The validation is cheap and happens at the service boundary before any proving work begins.

In short: fail fast at the RPC layer with an actionable hint, rather than failing deep inside the prover after significant compute has already been spent.

### LINKED ISSUE #346: bug: the detailed error is not transmitted as the response of a failed gRPC call

joanestebanr:
## Summary

This PR adds two independent safety features:

### Fix: #343 Detect broken import/unclaim pairs after range reduction 

When the proposer reduces the `end_block` of a certificate (range reduction), it is possible that an `imported_bridge_exit` falls within the new range while its matching `unclaim` does not. This would cause the import to appear as valid in the proof even though it was intentionally cancelled.

A new `validate_no_broken_pairs` check is added in `aggchain-proof-service` that, whenever the proposer returns a different `end_block` than the one originally requested, verifies that no `(imported_bridge_exit, unclaim)` pair is split by the new boundary. Pairs are matched by `global_index` in chronological order (k-th import ↔ k-th unclaim). If a broken pair is found, proof generation is aborted with a descriptive error.

### Fix: #346 full error chain returned to gRPC clients 

Previously, gRPC error responses only included the top-level error message (`error.to_string()`), silently dropping all nested cause messages. A new `format_error_chain` helper walks the full `source()` chain and concatenates all messages, so clients now receive the complete error description.

### Chore:  IGNORE_VKEY_MISMATCH environment variable to be able to develop/debug locally

Adds an opt-in escape hatch for development/testing environments: when `IGNORE_VKEY_MISMATCH` is set, vkey hash and range vkey commitment mismatches between the on-chain op-succinct config and the local ELF are logged as warnings instead of returning an error, allowing the server to continue running. The mismatch is still clearly surfaced in the logs. This flag should **never** be set in production.

## Changes

- `aggchain-proof-service/src/validation.rs` — new module with `validate_no_broken_pairs` and unit tests
- `aggchain-proof-service/src/service.rs` — calls the validation after receiving a proposer response with a reduced end block
- `aggchain-proof-service/src/error.rs` — new `BrokenImportUnclaimPair` error variant
- `aggchain-proof-builder/src/lib.rs` — `IGNORE_VKEY_MISMATCH` support for aggregation vkey hash and range vkey commitment checks
- `aggkit-prover/src/rpc.rs` — `format_error_chain` helper; full error chain now returned to gRPC clients
- `proposer-client/src/rpc/mod.rs` — minor error wrapping fix
- `scripts/generate_config_based_kurtosis_to_run_locally.sh` — new helper script for local Kurtosis-based testing

Fixes #343, #346


crates/aggchain-proof-service/src/validation.rs
```diff
@@ -0,0 +1,325 @@
...
+        let import_events = imports_by_global_index
+            .get(global_index)
+            .map(Vec::as_slice)
+            .unwrap_or(&[]);
+
+        for (k, &(unclaim_block, _)) in unclaim_events.iter().enumerate() {
+            let Some(&(import_block, _)) = import_events.get(k) else {
+                // No matching import for this unclaim position; not a split.
+                continue;
+            };
```
hadjiszs: can we end up in this `else`? Having one unclaim without a corresponding import in the block range that we are processing leads to an error normally no?

crates/aggchain-proof-service/src/validation.rs
```diff
@@ -0,0 +1,325 @@
...
+/// Validates that a proposer's reduced `new_end_block` does not split any
+/// (imported_bridge_exit, unclaim) pair.
+///
+/// The k-th unclaim for a given `global_index` (sorted by `(block_number,
+/// log_index)`) is matched to the k-th import with the same `global_index`
+/// (also sorted by `(block_number, log_index)`). A pair is broken when the
+/// import falls within the new block range but its matching unclaim does not,
+/// meaning the import would appear as valid in the proof even though it was
+/// intended to be cancelled.
+pub(crate) fn validate_no_broken_pairs(
```
hadjiszs, line 24: I asked claude to try to make this function more readable

it suggested something like below (not tested nor properly reviewed), wdyt? Happy to review another version if you find better 👌
but basically trying to have dedicated structs if possible, and splitting this function into "build pairs into well-suited data structures" followed by "check the pairs"

```rust
#[derive(Debug)]
struct ImportUnclaimPair {
    global_index: U256,
    import_block: u64,
    unclaim_block: u64,
}

pub(crate) fn validate_no_broken_pairs(
    imports: &[ImportedBridgeExitWithBlockNumber],
    unclaims: &[UnclaimWithBlockNumber],
    new_end_block: u64,
) -> Result<(), Error> {
    for ImportUnclaimPair {
        global_index,
        import_block,
        unclaim_block,
    } in pair_imports_with_unclaims(imports, unclaims)
    {
        if import_block <= new_end_block && unclaim_block > new_end_block {
            return Err(Error::BrokenImportUnclaimPair {
                global_index,
                import_block,
                unclaim_block,
                new_end_block,
            });
        }
    }
    Ok(())
}

/// Pairs the k-th import event for each `global_index` with the k-th unclaim
/// for the same `global_index`, both in (block_number, log_index) order.
/// Excess events on either side are dropped — they correspond to imports
/// not yet cancelled, or unclaims targeting imports outside this witness.
fn pair_imports_with_unclaims(
    imports: &[ImportedBridgeExitWithBlockNumber],
    unclaims: &[UnclaimWithBlockNumber],
) -> Vec<ImportUnclaimPair> {
    #[derive(Default)]
    struct EventLists {
        imports: Vec<(u64, u64)>,
        unclaims: Vec<(u64, u64)>,
    }

    let mut by_gi: HashMap<U256, EventLists> = HashMap::new();

    for ev in imports {
        by_gi
            .entry(ev.global_index.into())
            .or_default()
            .imports
            .push((ev.block_number, ev.log_index));
    }

    for ev in unclaims {
        by_gi
            .entry(ev.global_index)
            .or_default()
            .unclaims
            .push((ev.block_number, ev.log_index));
    }

    by_gi
        .into_iter()
        .flat_map(|(global_index, mut events)| {
            events.imports.sort_unstable();
            events.unclaims.sort_unstable();

            events
                .imports
                .into_iter()
                .zip(events.unclaims)
                .map(move |((import_block, _), (unclaim_block, _))| ImportUnclaimPair {
                    global_index,
                    import_block,
                    unclaim_block,
                })
        })
        .collect()
}
```

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -541,16 +541,25 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
-                        "Mismatch between the aggregation proof public values - retrieved from \
-                         the contracts: {retrieved_from_contracts:?}, received with the proof: \
-                         {:?}",
-                        aggregation_proof_public_values
-                    );
-                    return Err(Error::MismatchAggregationProofPublicValues {
-                        expected_by_contract: Box::new(retrieved_from_contracts),
-                        expected_by_verifier: Box::new(aggregation_proof_public_values.clone()),
-                    });
+                    if std::env::var("IGNORE_VKEY_MISMATCH").is_ok() {
```
hadjiszs, line 548: the check isn't only about vkeys here but on all the public values of the proof received from the op-succinct proposer. To check that they are the same as the ones expected by the L1 verifier (e.g., prev / new blocks, l1 head etc)

just to be sure I understand the motivation, this check is a problem with the `aggkit-prover` run in which way and with which setup?

hadjiszs (commented):
Thanks! Here is a first pass of review

crates/aggchain-proof-builder/src/lib.rs
joanestebanr: I introduce that to be able to run `aggkit-prover` even that is going to fail to return a proof becuase is not going to pass the verify of the proof locally. This allow to start the server and attend gRPC requests... 
If you think that can be dangerous in some way I'll remove it. WDTY?

crates/aggchain-proof-service/src/validation.rs
joanestebanr: Yes, this `else` is reachable and it's not an error. The inputs only contain the events within the range being proven. An unclaim cancels a previous import, but that import may have been claimed in an *earlier, already-proven* range — so it's not part of this witness. In that case there are more unclaims than imports for the `global_index`, and the trailing unclaim has no positional import → we hit the `else`.

This is legitimate: there's no in-range pair to split (the import is already finalized in a prior proof), so skipping it is correct. The core's `filter_values` in `verify_constrained_claims` also tolerates an unclaim without an in-range matching claim (it just removes nothing), so it doesn't error either.

crates/aggchain-proof-service/src/validation.rs
```diff
@@ -0,0 +1,264 @@
...
+            .get(global_index)
+            .map(Vec::as_slice)
+            .unwrap_or(&[]);
+
+        for (k, &unclaim_block) in unclaim_blocks.iter().enumerate() {
+            let Some(&import_block) = import_blocks.get(k) else {
+                // No matching import for this unclaim position; not a split.
+                continue;
+            };
+
```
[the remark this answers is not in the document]
joanestebanr: Already addressed in commit 52f5d2b: the per-`global_index` vectors now store `(block_number, log_index)` tuples and are sorted by the full tuple, so events within the same block are ordered by `log_index` and the k-th pairing is deterministic.

crates/aggchain-proof-service/src/validation.rs
```diff
@@ -0,0 +1,264 @@
...
+    let mut unclaims_by_global_index: HashMap<U256, Vec<u64>> = HashMap::new();
+    for unclaim in unclaims {
+        unclaims_by_global_index
+            .entry(unclaim.global_index)
+            .or_default()
+            .push(unclaim.block_number);
+    }
+    for blocks in unclaims_by_global_index.values_mut() {
+        blocks.sort_unstable();
+    }
```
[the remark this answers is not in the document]
joanestebanr: Same as the imports case — already addressed in commit 52f5d2b: unclaims are now stored as `(block_number, log_index)` tuples and sorted by the full tuple, so multiple unclaims for the same `global_index` in the same block are ordered by `log_index` and the k-th pairing is deterministic.

crates/aggchain-proof-service/src/validation.rs
joanestebanr: Thanks! Good suggestion — I went ahead and applied essentially this exact refactor in `refactor(validation): split pairing from broken-pair check`.

The current version now has:
- a dedicated `ImportUnclaimPair` struct,
- `validate_no_broken_pairs` reduced to just iterating the pairs and checking the broken-pair condition,
- the pairing logic extracted into `pair_imports_with_unclaims` (with the `EventLists` helper).

Only differences vs. your snippet are cosmetic: full-word variable names (`imported_bridge_exits`, `events_by_global_index`) to match the repo style, and slightly reworded doc comments. Let me know if you'd like anything else tweaked 👌

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -627,30 +636,49 @@ fn validate_op_succinct_config_keys(
     aggregation_vkey: &SP1VerifyingKey,
     expected_range_vkey_commitment: &Digest,
 ) -> Result<(), Error> {
+    let ignore_mismatch = std::env::var("IGNORE_VKEY_MISMATCH").is_ok();
```
hadjiszs, line 639: nit: otherwise `IGNORE_VKEY_MISMATCH=false` would enable it

```suggestion
    let ignore_mismatch = prover_utils::from_env_or_default("IGNORE_VKEY_MISMATCH", false);
```

crates/aggchain-proof-builder/src/lib.rs
hadjiszs: ok I think thats fine, if this is set in production by mistake the worst case will be failure on L1 settlement with aggchain hash mismatch, should be fine then

hadjiszs (approved):
lgtm 👍 

nit: we may get some issues if aggsender submits some unclaim that are prior to the `last_proven_block`, due to the order in which we sort and filter, but I guess aggsender already enforce that when generating its request, so I'm fine not bothering too much there

PR #348, opened 2026-05-15
chore: remove useless gnark dep

hadjiszs:
Remove unnecessary gnark dependency along with the download of circuit on the docker build


hadjiszs:
included in #349

PR #364, opened 2026-07-20
fix(deps): switch insta to crates.io

hadjiszs:


Ekleog-Polygon (dismissed):

Ekleog-Polygon (approved):
Weird that less-clean code would pass but not cleaner code, but 🤷 let's get that fix in

PR #369, opened 2026-09-21
feat(aggkit-prover): review+iterate on the noop recovery program

hadjiszs:
Review with suggested changes on top of 
- #367

## Why

On Bali, #367 as it stands cannot settle the recovery certificate: its aggchain params still carry the L2 pre-root, and the noop program is proven by the SP1 mock prover. It also overloads the word "mock" and adds a block-range check that belongs to aggkit.

### 1. The pre-root comes from L1

The program committed the params derived from the witness, with a pre-root rebuilt from L2 data. That is exactly the value that mismatches, so the `Aggchain hash mismatch` stayed.

Now: the noop program commits the `AggchainProofPublicValues` it is given. The builder builds the witness as before and, for `noop` only, recomputes `aggchain_params` with the root of `getL2Output(latestOutputIndex())`. It refuses a request not anchored at that output. The program depends only on `unified-bridge` and `sp1-zkvm`, so its vkey stops moving with the core crate; `MOCK_VKEY` is gone, the vkey is derived from the ELF.

### 2. Program and prover are two independent config axes

The new ELF was used only when the prover was `mock-prover`. A mock prover gives an SP1 mock proof, and the agglayer on Bali verifies the aggchain proof for real, so that certificate could never settle. Kurtosis missed it because everything there is mock, L1 verifier included.

Now: `program = "standard" | "noop"` next to `primary-prover`. What is proven and how it is proven are separate choices, all four combinations are explicit. Bali needs `noop` + `network-prover`. The selector follows the program.

### 3. `noop`, not `mock`

"mock" already means an SP1 mock proof (`mock-prover`). Reusing it for the new program made "mock aggchain proof" mean two opposite things: a fake proof of the real program, or a proof of the empty program. `noop` is the program. `mock` stays the SP1 proof mode. Renamed everywhere: crate, ELF path, constants, selector, CLI flags.

### 4. No safe-block guard in the prover

Choosing the block range and when to prove it is the aggsender's job: aggkit is the orchestrator and already has the knobs (`AdjustBlockRange`, `BridgeL2Sync.BlockFinality`). The prover should have no say on proving timing. It also cannot do it well: the aggsender signs the end block, so the prover can only reject, and with aggkit's default `LatestBlock` finality it would reject every optimistic certificate, the recovery one included. If wanted, this belongs in aggkit. Removed with its trait, tests and dev-dependencies.

## Runbook (in the README)

1. `aggkit-prover vkey --noop`, `aggkit-prover vkey-selector --noop`.
2. The L2 local exit root at the last settled block must equal L1's; otherwise `BackwardLET` / `ForwardLET` first.
3. Aggchain manager: `addOwnedAggchainVKey(0xFFFF0001, <vkey>)`, then `disableUseDefaultVkeysFlag()`.
4. `program = "noop"`, regular prover type, restart the prover.
5. Drop the aggsender's cached aggchain proof (resync its DB) so it calls the prover again.
6. Settled: `enableUseDefaultVkeysFlag()`, `program = "standard"`, restart.

## Testing

- fmt with the CI nightly, clippy `-D warnings`, nextest on the touched crates.
- New tests: the params override matches the program's encoding; the committed noop ELF commits exactly what it is given; selector value.
- Not run end-to-end against a real verifier yet.

## Before merging

- The noop ELF was built locally. Rebuild it with the reproducible build (`AGGLAYER_ELF_BUILD=update cargo build`) and register the vkey printed from that binary.
- Set `program = "standard"` back as soon as the recovery certificate settles.


crates/aggchain-proof-builder/build.rs
```diff
@@ -3,17 +3,17 @@ use agglayer_elf_build::ProgramBuilder;
...
-    let mock_elf_path = ProgramBuilder::new("crates/aggchain-proof-mock-program")?
+    let noop_elf_path = ProgramBuilder::new("crates/aggchain-proof-noop-program")?
         .cached_elf_path(format!(
-            "{manifest_dir}/elf/mock/riscv64im-succinct-zkvm-elf"
+            "{manifest_dir}/elf/noop/riscv64im-succinct-zkvm-elf"
         ))
         .run()?;
-    println!("cargo::rustc-env=AGGLAYER_MOCK_ELF_PATH={mock_elf_path}");
-    eprintln!("MOCK_ELF_PATH={mock_elf_path}");
+    println!("cargo::rustc-env=AGGLAYER_NOOP_ELF_PATH={noop_elf_path}");
```
krlosMata, line 15: Blocker: the committed noop ELF isn't reproducible. Rebuilding it (`AGGLAYER_ELF_BUILD=update cargo build`) changes its bytes and its vkey — committed `0x27dd4d13…` vs reproducible `0x41f62272…`. So the runbook step-1 vkey (read from the committed binary) is wrong. Rebuild via the reproducible build and register the vkey from that before merge. Note `check_elf` is `SKIPPED` on `pull_request` (push/merge-queue only), so it won't catch this here — worth a `workflow_dispatch` run first.

krlosMata (commented):
Approach LGTM — this fixes what #367 couldn't: the settled-pre-root mismatch and the SP1-mock-proof-rejected-by-a-real-verifier problem. Verified locally on this branch: `fmt` (CI nightly), `clippy -D warnings`, `nextest` (50 passed), doc tests and `--locked` all green.

Inline: one blocker (stale noop ELF) and one should-fix (recovery still hits the failing RPC). A few more:

**Safe-block guard removal (your call).** Reasonable to move range/timing to the aggsender, and the "it'd reject the recovery cert under `LatestBlock`" point is correct. Flagging only because it was an intentionally-requested guard — fine to drop if it's re-homed in aggkit; just want it recorded.

**nit — `noop` + `mock-prover` is still a valid combo** that can't settle on a real verifier. A `warn!`/validation when `program = noop` with `mock-prover` would prevent the "worked in Kurtosis, fails on Bali" trap the README calls out.

**nit — test coverage.** `params_with_the_same_pre_root_match_the_program_encoding` only asserts the no-op case (pre-root unchanged → equals the normal encoding). Consider also asserting a *different* (L1) pre-root yields the expected distinct hash — the actual recovery path.

Nice: vkey derived from the ELF, noop program limited to `unified-bridge`+`sp1-zkvm` (stable selector), the `noop_elf_commits_the_given_public_values` execution test, and the runbook.

crates/aggchain-proof-builder/src/lib.rs
```diff
@@ -644,18 +649,33 @@ impl<ContractsClient> AggchainProofBuilder<ContractsClient> {
...
-                        .proof
-                        .clone()
-                        .try_as_compressed()
-                        .ok_or(Error::GeneratedProofIsNotCompressed)?;
-                    stdin.write_proof(*aggregation_proof, aggregation_vkey.vk.clone());
+                match noop_pre_root {
+                    // The noop program commits the public values as given, so
+                    // the pre-root can be the one the contract hashes, which
+                    // the L2 no longer has after the reorg.
+                    Some(l1_pre_root) => {
```
krlosMata: Should-fix: in noop mode only `public_values` is written here, but `retrieve_chain_data` still ran `get_prev_l2_block_sketch` / `get_new_l2_block_sketch` (+ the static calls) and we discard them. That pre-block sketch is the exact call that fails with `MismatchedStateRoot` during the reorg being recovered from (seen live on Bali:84). Consider skipping witness/sketch building in noop mode — fetch only what `public_values` needs (prev/new LER, imported-bridge commitment, L1 info root) — so recovery doesn't depend on the L2 node serving a clean pre-block sketch.

